PR #2, opened 2022-10-13
Fix link to a changelog

ligurio:


Zac-HD:
Thanks for the PR!  I actually just pushed a slightly different fix along with the general "make a release" patch, so closing this despite appreciating it🙂

PR #20, opened 2023-04-25
fixing arg name etype in format_exception

### LINKED ISSUE #16: Failures not reported on python >= 3.10
`hy.py` fails to complete its task of collecting an actual failure because it fails itself within `_run_test_on()` for python 3.10 or newer like so:

    traceback.format_exception(etype=type(e), value=e, tb=tb)
    TypeError: format_exception() got an unexpected keyword argument 'etype'

I think this is because the signature of `traceback.format_exception()` changed from [3.9](https://docs.python.org/3.9/library/traceback.html#traceback.format_exception) to [3.10](https://docs.python.org/3.10/library/traceback.html#traceback.format_exception) to make the first argument positional only.

Cheukting:
arg name `etype` has changed in Python 3.10, use positional arg instead.

Closes #16

### DIFF
--- a/src/hypofuzz/hy.py
+++ b/src/hypofuzz/hy.py
@@ -272,7 +272,7 @@ class FuzzProcess:
             filename, lineno, *_ = traceback.extract_tb(tb)[-1]
             data.interesting_origin = (type(e), filename, lineno)
             data.extra_information.traceback = "".join(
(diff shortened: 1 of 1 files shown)


Zac-HD (approved):
🎉

Issue #3, opened 2022-10-26
Database-centric architecture for communication, persistence, and autoscaling


Zac-HD:
# The Status Quo

This is going to be substantial architecture overhaul, so let's start with how things currently work: a HypoFuzz run has three basic parts:

1. The Hypothesis database, a key-value store of failing and covering examples we've seen in previous runs (or other workers in this run)
2. The worker processes, which spend their time executing test cases for a variety of test functions, plus some 'meta' scheduling
3. The dashboard process, which serves a webpage showing how things are going, based on information sent by the workers over http.

In the current design, this is fundamentally a run-it-on-one-box kind of system: the tests are divided up between workers at startup time (or maybe run on every worker concurrently; the workers are fine with this though the dashboard isn't), and while the workers can reload the previous examples everything else is as if it were the first run ever - with some hit to efficiency and the clarity of statistics.

**Goal:** support a system where workers can come and go, for example to soak up idle CPU time as a low-priority autoscaling group on a cluster, and the fuzzing system overall keeps humming along.

# Solution: lean on the database

If our problem is that information is neither persisted nor well distributed, let's solve that with [the Hypothesis database](https://hypothesis.readthedocs.io/en/latest/database.html)!  This is a very simple key-value store where keys are bytestrings and values are sets of bytestrings, with create/read/delete operations.  The most common implementation is on the user's local filesystem, but there's also a Redis backend and it's trivial to write more.

What problems does this solve, and create?

- ✨ Workers could write metadata to the database (in some disjoint keyspace), meaning that the dashboard could show information regardless of whether a worker is currently running - it'd just be a view over the database (generally good design principle!), and update by polling at whatever frequency we wanted.
  - ✨ We no longer need any HTTP traffic between components of the system; subtracting parts is underrated.
  - ✨ If we had two separate systems ("partition tolerance"), we can just merge the databases (via e.g. `MultiplexedDatabase`) and keep going
  - 🚧 We *have to* handle stale data, including from runs that diverged or never even had a common prefix.  This is basically fine; "keep the examples from everything and discard all the metadata" is totally valid and anything fancier is a bonus.  We'll probably try to construct a 'best guess' metadata though, e.g. keeping the longest.  
[...]

Issue #5, opened 2022-10-26
Report coverage stability in the dashboard


Zac-HD:
We should also report coverage stability fraction, including a rating stable (100%), unstable (85%--100%), or serious problem (<85%"); and explain the difference between stability (=coverage) and flakiness (=outcome).  Stability is mostly an efficiency thing; flakiness means your test is broken.

This mostly requires *measuring* both of these on the backend and then plumbing the data around, it's not hugely involved.

tybug:
A silly question: what *is* coverage stability? I'd guess some jaccard similarity-derived metric of coverage between all runs of the same input, such that "100%" is deterministic and anything else is not?

Zac-HD:
Fraction of (repeated) inputs for which we observe identical coverage each time; it's pretty easy to measure.  Concretely I'd replay from our seed pool at a diminishing rate (say 1% of executions, halve the rate each time through the pool, over-schedule new pool entries to catch up, replay entries twice occasionally as a cache-hit heuristic), so we get some decent confidence fairly promptly but with low overhead over time.

We're really just reporting an observation here rather than estimating some latent property, partly because it's unclear what that would be.  e.g. conceptually we might have different flake-rates per input, may or may not care about that, etc.  A particularly common case is that we observe different coverage _only the first time_ we play some input; it's ambiguous whether we actually want such a seed in our pool since mutants are less likely to help.

PR #41, opened 2024-11-16
Migrate dashboard to polling the db

tybug:
#3 

Mostly checking that this is the right path forward, and getting a feel for the architecture of the codebase. `msgpack` isn't a hard dependency, we could easily roll our own b64-json-encoded serialization. (but mspack is right there...)

I've noticed coverage dip when restarting the process, after it replays its covering examples. Possibly our criteria for saving inputs isn't synced with the coveragepy coverage, so replaying doesn't get us fully back there. Or we have an actual bug, not sure.

image

P.S. running `hypothesis fuzz` and having fuzzing *just work* feels so incredibly magical.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setuptools.setup(
         "psutil >= 3.0.0",
         "pytest >= 6.0.1",
         "requests >= 2.24.0",
+        "msgpack >= 1.1.0",
     ],
     extras_require={
         "pytrace": [
--- a/src/hypofuzz/dashboard.py
+++ b/src/hypofuzz/dashboard.py
@@ -4,9 +4,10 @@ import atexit
 import datetime
 import os
 import signal
-from typing import List, Tuple
+from typing import List
 
 import black
+import msgpack
 import dash
 import flask
 import plotly.express as px
@@ -14,6 +15,7 @@ import plotly.graph_objects as go
 from dash import dcc, html
 from dash.dependencies import Input, Output
 from hypothesis.configuration import storage_directory
+from hypothesis import settings
 
 from .patching import make_and_save_patches
 
@@ -34,23 +36,25 @@ else:
     flask_cors.CORS(app)
 
 
-@app.route("/", methods=["POST"])  # type: ignore
-def recv_data() -> Tuple[str, int]:
-    data = flask.request.json
-    if not isinstance(data, list):
-        data = [data]
-    for d in data:
-        add_data(d)
-    return "", 200
+def poll_database() -> None:
+    global DATA_TO_PLOT
 
+    data = []
+    for key in settings.default.database.fetch(b"hypofuzz-test-keys"):
+        data += [
+            msgpack.loads(v)
+            for v in settings.default.database.fetch(key + b".hypofuzz.metadata")
+        ]
+    if not data:
+        return
+    data = sorted(data, key=lambda data: data["elapsed_time"])
 
-def add_data(d: dict) -> None:
     if not LAST_UPDATE:
         del DATA_TO_PLOT[0]
-    DATA_TO_PLOT.append(
-        {k: d[k] for k in ["nodeid", "elapsed_time", "ninputs", "branches"]}
-    )
-    LAST_UPDATE[d["nodeid"]] = d
+
+    DATA_TO_PLOT = data
+    for d in data:
+        LAST_UPDATE[d["nodeid"]] = d
 
 
 external_stylesheets = ["https://codepen.io/chriddyp/pen/bWLwgP.css"]
@@ -98,6 +102,8 @@ def try_format(code: str) -> str:
     [Input("url", "pathname")],
 )
 def display_page(pathname: str) -> html.Div:
+    poll_database()
+
     # Main page
     if pathname == "/" or pathname is None:
         return html.Div(
@@ -200,6 +206,8 @@ FIRST_FAILED_AT = {}
     [Input("interval-component", "n_intervals"), Input("xaxis-state", "n_clicks")],
 )
 def update_graph_live(n: int, clicks: int) -> object:
+    poll_database()
+
     fig = px.line(
         DATA_TO_PLOT,
         x="ninputs",
--- a/src/hypofuzz/hy.py
+++ b/src/hypofuzz/hy.py
@@ -7,6 +7,7 @@ import time
 import traceback
 from random import Random
 from typing import Any, Callable, Dict, Generator, List, Optional, Union
+import msgpack
 
 from hypothesis import settings
 from hypothesis.core import (
@@ -37,8 +38,6 @@ except ImportError:
     record_pytrace = None
 
 Report = Dict[str, Union[int, float, str, list, Dict[str, int]]]
-
-UNDELIVERED_REPORTS: List[Report] = []
 
 
 @contextlib.contextmanager
@@ -122,6 +121,7 @@ class FuzzProcess:
         self._test_fn = test_fn
         self.__stuff = stuff
         self.nodeid = nodeid or test_fn.__qualname__
+        self.database_key = database_key
 
         # The seed pool is responsible for managing all seed state, including saving
         # novel seeds to the database.  This includes tracking how often each branch
@@ -143,14 +143,23 @@ class FuzzProcess:
         # 1000 consecutive examples without new coverage, and then switch to mutation.
         self._early_blackbox_mode = True
 
+        metadata = [
+            msgpack.loads(v) for v in settings.default.database.fetch(self.metadata_key)
+        ]
+        if metadata:
+            latest = sorted(metadata, key=lambda data: data["elapsed_time"])[-1]
+            self.ninputs = latest["ninputs"]
+            self.elapsed_time = latest["elapsed_time"]
+
         # We batch updates, since frequent HTTP posts are slow
         self._last_post_time = self.elapsed_time
+        self._reports: List[Report] = []
 
     def startup(self) -> None:
         """Set up initial state and prepare to replay the saved behaviour."""
-        assert self.ninputs == 0, "already started this FuzzProcess"
-        # Report that we've started this fuzz target, and run zero examples so far
-        self._report_change(self._json_description)
+        if self.ninputs == 0:
+            self._reports.append(self._json_description)
+            self._report_changes()
         # Next, restore progress made in previous runs by loading our saved examples.
         # This is meant to be the minimal set of inputs that exhibits all distinct
         # behaviours we've observed to date.  Replaying takes longer than restoring
@@ -197,7 +206,6 @@ class FuzzProcess:
             self._replay_buffer.extend(self.pool.fetch())
 
         # seen_count = len(self.pool.arc_counts)
-
         # Run the input
         result = self._run_test_on(
             ConjectureData(
@@ -228,9 +236,8 @@ class FuzzProcess:
                     shrinker.shrink_target,
                     collector=record_pytrace(self.nodeid),
                 )
-            UNDELIVERED_REPORTS.append(self._json_description)
-            self._report_change(UNDELIVERED_REPORTS)
-            del UNDELIVERED_REPORTS[:]
+            self._reports.append(self._json_description)
+            self._report_changes()
 
         # Consider switching out of blackbox mode.
         if self.since_new_cov >= 1000 and not self._replay_buffer:
@@ -330,13 +337,12 @@ class FuzzProcess:
             self.since_new_cov = 0
         else:
             self.since_new_cov += 1
-        if 0 in (self.since_new_cov, self.ninputs % 100):
-            UNDELIVERED_REPORTS.append(self._json_description)
+        if 0 in (self.since_new_cov, self.ninputs % 100) and not self._replay_buffer:
+            self._reports.append(self._json_description)
 
         self.elapsed_time += time.perf_counter() - start
-        if UNDELIVERED_REPORTS and (self._last_post_time + 10 < self.elapsed_time):
-            self._report_change(UNDELIVERED_REPORTS)
-            del UNDELIVERED_REPORTS[:]
+        if self._last_post_time + 10 < self.elapsed_time:
+            self._report_changes()
 
         if self.elapsed_time > self.stop_shrinking_at:
             raise HitShrinkTimeoutError
@@ -344,8 +350,18 @@ class FuzzProcess:
         # The shrinker relies on returning the data object to be inspected.
         return data.as_result()
 
-    def _report_change(self, data: Union[Report, List[Report]]) -> None:
-        """Replace this method to send JSON data to the dashboard."""
+    @property
+    def metadata_key(self):
+        return self.database_key + b".hypofuzz.metadata"
+
+    def _report_changes(self) -> None:
+        # don't report while replaying the corpus, since coverage won't be
+        # up to its full value during this process.
+        if self._replay_buffer:
+            return
+        while self._reports:
+            report = self._reports.pop(0)
+            settings.default.database.save(self.metadata_key, msgpack.dumps(report))
 
     @property
     def _json_description(self) -> Report:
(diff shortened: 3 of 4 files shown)


src/hypofuzz/hy.py
```diff
@@ -330,22 +337,31 @@ def _run_test_on(
...
-            UNDELIVERED_REPORTS.append(self._json_description)
+        if 0 in (self.since_new_cov, self.ninputs % 100) and not self._replay_buffer:
+            self._reports.append(self._json_description)
 
         self.elapsed_time += time.perf_counter() - start
-        if UNDELIVERED_REPORTS and (self._last_post_time + 10 < self.elapsed_time):
-            self._report_change(UNDELIVERED_REPORTS)
-            del UNDELIVERED_REPORTS[:]
+        if self._last_post_time + 10 < self.elapsed_time:
+            self._report_changes()
```
Zac-HD, line 345: Given the changes to report delivery, in particular that we can naturally batch http requests but can't batch database writes, I think we should make a couple of changes here.

- report changes immediately, rather than waiting 10s
- move the actual database logic to a background thread, so that the fuzzing worker just drops the message in a queue and moves on.  
  - Plausibly this should take the form of an `InBackgroundThreadDatabase(db)` wrapper which can live in Hypothesis itself?  Fuzzing can't be the only usecase, and doing some smart stuff like issuing writes before deletes in case of shutdown would be pretty easy to handle in one place.  ("database is only ever best-effort" is such a nice invariant)
- for the `self.ninputs % 100 == 0` case, we should track the previous report and if nothing has changed, drop it from the database.  This is mostly a win for the plotting code, which gets unhappy with very large numbers of points, but also analysis in general.

unrelatedly it'd be nice to have a stats-as-json endpoint that users can hit to get all the raw data we're displaying, in case they want to slice and maybe display it for themselves 🤔

Zac-HD (commented):
> I've noticed coverage dip when restarting the process, after it replays its covering examples. Possibly our criteria for saving inputs isn't synced with the coveragepy coverage, so replaying doesn't get us fully back there.

My best guess is that your test is just nondeterministic with respect to coverage, such that you _literally ever_ covered branches which didn't get executed on replay.  See https://github.com/Zac-HD/hypofuzz/issues/5; some of my notes below assume that stable coverage is ~default which seems likely to be both false and a reasonable design position (!!).  If it's _very_ rare we might need to reconsider, but we can start conservative and reassess with data from our own projects.

> P.S. running hypothesis fuzz and having fuzzing just work feels so incredibly magical.

💖 

that's the goal!  Above all else, it has to be _so freaking easy_ that developers actually adopt fuzzing - that's the only way to drive 100x adoption from the (decent!) status quo.

---------------

Also nice here, let's add a `hypothesis fuzz --dashboard-only` CLI arg which does the obvious?

Zac-HD:
Also, uh, we might need to update `hypofuzz` pretty soon to work with the IR; that won't be a huge job and notably should _not_ block dropping buffer support on the Hypothesis side, but fuzzing will need an update 😁

src/hypofuzz/dashboard.py
```diff
@@ -34,23 +36,25 @@
...
-    return "", 200
+def poll_database() -> None:
+    global DATA_TO_PLOT
 
+    data = []
+    for key in settings.default.database.fetch(b"hypofuzz-test-keys"):
+        data += [
+            msgpack.loads(v)
+            for v in settings.default.database.fetch(key + b".hypofuzz.metadata")
+        ]
```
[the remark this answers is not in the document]
tybug: yeah, my .hypothesis is already 12mb locally. We can probably just store less in the metadata entries and keep only the latest full entry in `.hypofuzz.latest`.

setup.py
```diff
@@ -43,6 +43,7 @@ def local_file(name: str) -> str:
         "psutil >= 3.0.0",
         "pytest >= 6.0.1",
         "requests >= 2.24.0",
+        "msgpack >= 1.1.0",
```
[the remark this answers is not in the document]
tybug: oops, of course!

tybug:
I'm not sure there's anything easy in hypofuzz to migrate to the IR? Saving failing examples to the hypothesis db needs to wait until the hypothesis db is on the ir, and migrating the mutator to the ir is itself a major task.

The thought dumps here are much appreciated, as I'm still in the 'aligning expectations' phase of working on hypofuzz <3.

src/hypofuzz/hy.py
```diff
@@ -330,22 +337,31 @@ def _run_test_on(
...
-        """Replace this method to send JSON data to the dashboard."""
+    @property
+    def metadata_key(self):
+        return self.database_key + b".hypofuzz.metadata"
+
+    def _report_changes(self) -> None:
+        # don't report while replaying the corpus, since coverage won't be
+        # up to its full value during this process.
+        if self._replay_buffer:
+            return
```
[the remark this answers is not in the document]
tybug: > I'd be inclined to replay anyway during this period, and just insert an extra key to distinguish the stage.

we actually already were doing this, whoops! After thinking about it I agree hiding this stage from the plot is a bad idea.

src/hypofuzz/database.py
```diff
@@ -0,0 +1,35 @@
...
+from hypothesis.database import BackgroundWriteDatabase, ExampleDatabase
+
+
+def metadata_key(key):
+    return key + b".hypofuzz.metadata"
+
+
+class MetadataDatabase(ExampleDatabase):
+    def __init__(self, db: ExampleDatabase) -> None:
+        self._db = db
```
tybug, line 14: this abstraction might be overkill...

src/hypofuzz/hy.py
```diff
@@ -142,15 +143,19 @@ def __init__(
         # After replay, we stay in blackbox mode for a while, until we've generated
         # 1000 consecutive examples without new coverage, and then switch to mutation.
         self._early_blackbox_mode = True
+        self._last_report: Report | None = None
 
-        # We batch updates, since frequent HTTP posts are slow
-        self._last_post_time = self.elapsed_time
+        metadata = db.fetch_metadata(self.database_key)
+        if metadata:
+            latest = sorted(metadata, key=lambda data: data["elapsed_time"])[-1]
```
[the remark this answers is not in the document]
tybug: what a nice trick! also cleaner to read, of course, but the O improvement is cute on its own.

src/hypofuzz/database.py
```diff
@@ -0,0 +1,35 @@
...
+        self._db.save(metadata_key(key), value)
+
+    def delete_metadata(self, key: bytes, value: bytes) -> None:
+        self._db.delete(metadata_key(key), value)
+
+    def fetch_metadata(self, key: bytes) -> Iterable[bytes]:
+        return [json.loads(v) for v in db.fetch(metadata_key(key))]
+
+
+db = MetadataDatabase(BackgroundWriteDatabase(settings.default.database))
```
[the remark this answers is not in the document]
tybug: I think we want a hard cutoff point where we create a database singleton, otherwise we have a thread.start overhead for every db access. But yeah we should be creating the db post-something.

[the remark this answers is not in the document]
Zac-HD: Yeah, maybe we wrap this in a `@functools.lru_cache(maxsize=1)` or something?  I agree we only want one per process, it's just a matter of when we create it.

src/hypofuzz/hy.py
```diff
@@ -331,21 +333,30 @@ def _run_test_on(
...
+        # notes in the future.
+        if (
+            self._last_report is not None
+            and report["branches"] == self._last_report["branches"]
+        ):
+            db.delete_metadata(
+                self.database_key, bytes(json.dumps(self._last_report), "ascii")
+            )
+
+        self._last_report = report
```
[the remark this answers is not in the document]
tybug: good catch for `since_new_cov`, we want to avoid dropping reports which discovered new coverage

Zac-HD:
Ah, you'll need to `tox -e deps` to update the pinned dependencies.

tybug:
thanks, first time using tox. And I guess I really need `python3.10 -m tox -e deps` to avoid dep changes from my default 3.12?

tests/test_e2e.py
```diff
@@ -31,15 +33,34 @@ def test_end_to_end(numprocesses, tmp_path):
     """An end-to-end test to start the fuzzer and access the dashboard."""
     test_fname = tmp_path / "test_demo2.py"
     test_fname.write_text(TEST_CODE, encoding="utf-8")
-    dash_proc = entrypoint._fuzz_impl(
-        numprocesses=numprocesses,
-        dashboard=True,
-        port=7777,
-        unsafe=False,
-        pytest_args=["-p", "no:dash", str(test_fname)],
+    process = subprocess.Popen(
```
tybug, line 36: I thought this might be better than trying to expose the dashboard process in the code. I think the alternative is to set `dash_proc` as a global somewhere, or maybe there's an easy way to return it that I'm not seeing.

deps/test-old.txt
```diff
@@ -5,59 +5,59 @@
...
+psutil==6.1.0             # via hypofuzz (setup.py)
+pyarrow==18.1.0           # via -r /Users/tybug/Desktop/Liam/coding/hypofuzz/deps/test.in
 pygments==2.18.0          # via rich
-pytest==7.4.4             # via -r deps/./test.in, -r deps/test-old.in, hypofuzz (setup.py), pytest-cov, pytest-xdist
-pytest-cov==5.0.0         # via -r deps/./test.in
-pytest-xdist==3.6.1       # via -r deps/./test.in
-python-dateutil @ git+https://github.com/dateutil/dateutil.git@296d419fe6bf3b22897f8f210735ac9c4e1cb796  # via -r deps/./test.in, pandas
+pytest==7.4.4             # via -r /Users/tybug/Desktop/Liam/coding/hypofuzz/deps/test.in, -r deps/test-old.in, hypofuzz (setup.py), pytest-cov, pytest-xdist
+pytest-cov==6.0.0         # via -r /Users/tybug/Desktop/Liam/coding/hypofuzz/deps/test.in
+pytest-xdist==3.6.1       # via -r /Users/tybug/Desktop/Liam/coding/hypofuzz/deps/test.in
```
tybug, line 45: is there a way to get tox to use a relative path? I guess this is the same thing we ran into in https://github.com/HypothesisWorks/hypothesis/pull/4178#discussion_r1845263735 huh?

tests/test_e2e.py
Zac-HD: `subprocess` seems reasonable to me 👍

deps/test-old.txt
Zac-HD: Haven't looked into it on either end yet; probably the quickest way forward is to do some `sed` on the files after writing.

Zac-HD:
Poked at some stuff, pick it out of https://github.com/Zac-HD/hypofuzz/commit/de1ccbfe8d8aab778392858dd27e006c23a8d82b as you wish 🙂

src/hypofuzz/__init__.py
```diff
@@ -1,4 +1,4 @@
 """Adaptive fuzzing for property-based tests using Hypothesis."""
 
-__version__ = "24.9.1"
+__version__ = "24.11.1"
```
Zac-HD, line 3: alas, 😅

setup.py
```diff
@@ -58,9 +57,11 @@ def local_file(name: str) -> str:
         "License :: OSI Approved :: GNU Affero General Public License v3",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.8",
```
Zac-HD, line 61: oops, should have updated `python_requires=">=3.8",` above too

docs-src/conf.py
Zac-HD, line 1: I think my cursed old git config messed up the line-endings here?  We should fix that...

src/hypofuzz/dashboard.py
```diff
@@ -13,11 +14,13 @@
...
 from dash import dcc, html
 from dash.dependencies import Input, Output
+from hypothesis import settings
 from hypothesis.configuration import storage_directory
 
+from .database import get_db
 from .patching import make_and_save_patches
 
-DATA_TO_PLOT = [{"nodeid": "", "elapsed_time": 0, "ninputs": 0, "branches": 0}]
+DATA_TO_PLOT = []
```
Zac-HD, line 23: We should test that the dashboard doesn't crash if run against an empty database.

Also we should open an issue to have the dashboard show a useful message in that case, to the effect of "The database ({database!r}) doesn't have contain fuzzing reports.  If you've run hypofuzz before, check that the database configured on the workers is the same as for the dashboard."  (after a ~5-10s timeout to avoid the freshly-started-popup problem).

src/hypofuzz/hy.py
```diff
@@ -346,7 +346,8 @@ def _run_test_on(
         return data.as_result()
 
     def _report(self, report: Report) -> None:
-        (db := get_db()).save_metadata(self.database_key, report)
```
tybug: I'm a fan of the walrus operator for "define in if guard, use in body", but I find this type of usage (define on first use) a bit gratuitous and unclear, idk. maybe I just need to get used to it 🙂

tybug:
re: [ci fail](https://github.com/Zac-HD/hypofuzz/actions/runs/12152828038/job/33889890439?pr=41), any objections to moving the flake8 linting to ruff? I think hypothesis is on ruff, and I'm guessing hypofuzz is old enough that it predates ruff having enough flake8 rules ported to be worth considering.

src/hypofuzz/hy.py
Zac-HD: I agree, this is less clear that I want the code to be.

src/hypofuzz/corpus.py
Zac-HD, line 1: my bad on line-endings again 😓

Zac-HD:
No objections at all to switching linting over - hypofuzz just predates ruff.  (we might also be able to configure `tox` to run all commands even if one fails?)

But let's ship this and then do a follow-up PR 🙂

src/hypofuzz/__init__.py
tybug: I have definitely been spoiled by never having to worry about bumping in hypothesis...

src/hypofuzz/dashboard.py
tybug: Thinking aloud about a future `test_dashboard.py`: define a `dashboard` fixture which starts and returns a `--dashboard-only` fixture. Tests can launch their own `--no-dashboard` process and then use the proposed dashboard-json-api to check for correctness of the fuzzer.

To the problem at hand: how to control the database used by a `--dashboard-only` process? I'd expect with a `--profile` option. Which means `--dashboard-only` needs to execute a pytest collection to pick up on any user-defined profiles? Does this all sound about correct?

(I'm thinking ahead here because the requested database-isolation across tests is difficult given that the hypothesis default is a shared filesystem db).

docs-src/conf.py
tybug: yep, `git add --renormalize .` should do it, at least whenever we're not concerned about ballooning the diff

src/hypofuzz/dashboard.py
Zac-HD: I think `--profile=...` is redundant with `-- --hypothesis-profile=...` (i.e. passing that as a pytest argument); we definitely have to execute a pytest collection step to execute the profile-defining code.

Ideally we'd also have a way to specify a particular database object (so we can use `InMemoryExampleDatabase` for testing); that would need to start a dashboard server within the current process.  Hmm.

tybug: Good point for `--hypothesis-profile`! Though we may want a convenience alias anyway.

My first thought is to continue running the database via multiprocessing in release, and have a `factory: type[Thread]  | type[Process] = Process` argument which we switch to threading in tests when sharing a db instance. But the api surface isn't quite the same (no `Thread.kill`)..

Zac-HD: I'm overall -0.5 on the convenience alias; on one hand it _is_ more convenient; but on the other I really want to push people to think about this as "take your pytest invocation and `s/pytest/hypothesis fuzz --/` and you're done"

tybug: yeah, that's fair.

PR #42, opened 2024-12-10
Move from flake8 to ruff

tybug:
Config largely copied from hypothesis.

Also includes a horrible commit of `git add --renormalize .` in https://github.com/Zac-HD/hypofuzz/pull/42/commits/3a86bbba104f54dfb790924cd4bcbacdc826423c. We could add that commit to `.git-blame-ignore-revs`, but I actually don't think it's worth the bloat of a new top-level file given the greenfield stage of the project.


Zac-HD (approved):
🙏

PR #43, opened 2024-12-12
Allow passing host when starting the dashboard

hoefling:
The reason for this PR is that I am often running the hypofuzz server in my homelab now and in order to access the dashboard from my workstation, I always need to apply this patch after installation :see_no_evil: Would it be acceptable to actually incorporate this as a cli argument? I intentionally kept the changes minimal, without adding IPv4/IPv6/FQDN validation, but if you find it useful, I can also add a custom `click` type instead of the plain `str`.

```diff
@@ -344,7 +344,7 @@ def patch_summary() -> flask.Response:
 
 
 def start_dashboard_process(
-    port: int, *, pytest_args: list, host: str = "localhost", debug: bool = False
+    port: int, *, pytest_args: list, host: str = "0.0.0.0", debug: bool = False
 ) -> None:
     global PYTEST_ARGS
     PYTEST_ARGS = pytest_args
```

### DIFF
--- a/src/hypofuzz/entrypoint.py
+++ b/src/hypofuzz/entrypoint.py
@@ -31,6 +31,13 @@ import psutil
     help="serve a live dashboard page without launching associated fuzzing processes",
 )
 @click.option(
+    "--host",
+    type=str,
+    default="localhost",
+    metavar="HOST",
+    help="Optional host for the dashboard",
+)
+@click.option(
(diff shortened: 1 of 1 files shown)


Zac-HD (approved):
I'm happy to take this patch - and glad you're finding Hypofuzz useful!

Issue #11, opened 2022-10-26
Pythia-style predictive fuzzing


Zac-HD:
We've got basic predictions of branches-to-next-new-branch and branches-to-first-bug in the dashboard, but we can do more with this:

- use [predictive fuzzing like Pythia](https://github.com/mboehme/pythia) for prioritization of tests to fuzz; currently using number of inputs since last discovery
- report expected runtime until next branch/bug in the dashboard
- clearly explain why residual risk estimation is awesome on the website
- allow user configuration of a cost per hour (per core?), and display cost estimates too
- report a global estimate aggregated across running tests (requires dashboard to know/estimate number of active workers)
- have a configurable threshold 'price per failure' at which to stop fuzzing

It would be really, really useful to collect lots of empirical measurements here in order to tune estimates and get around adaptive biases at least a little ([see paper](https://mboehme.github.io/paper/FSE21.pdf)).  On the other hand nobody likes telemetry.

tybug:
Additional work in this area from Böhme et al: [Extrapolating Coverage Rate in Greybox Fuzzing
](https://dl.acm.org/doi/pdf/10.1145/3597503.3639198)

CyberiaResurrection:
@Zac-HD , @tybug  , how about starting with an estimation of the _total_ number of branches that a given fuzzing campaign can reach?  Such as via one of the Chao estimators?

For example, 5908 branches reached on its own isn't too meaningful.   But 5908 branches reached out of an estimated (for instance) 6000 total is more meaningful - for some runs, that might be enough coverage.

I haven't dug into the code, so some or all of this may already be done, but above the current branch total, I think the basic Chao estimator only needs number of singletons and doubletons.

That also provides a jumping-off point for discovery probability or residual risk estimation, as outlined in the two papers linked.

tybug:
I would couch that `5908 / 6000` is not particularly useful if 5k of those lines were trivially run by the very first input! 

I believe that the latter paper uses singletons $f_1$ and doubletons $f_2$ in their estimator as well, so collecting those now and reporting the Chao1 estimate (plus a metric for "effective lines covered"; simply dropping coverage for the first 10 inputs may work fine) along the way may not be a bad idea.

CyberiaResurrection:
@tybug, I've been using Hypofuzz in a "run Hypofuzz until property violation found - fix violation - repeat" cycle.

[...]

PR #44, opened 2024-12-18
Implement Chao lower bound on number of branches

CyberiaResurrection:
As mentioned in https://github.com/Zac-HD/hypofuzz/issues/11#issuecomment-2537950837 , add singleton and doubleton counting to `Pool`.
After that, calculate the basic Chao bound on number of branches after the later of 10 inputs into a run, or previously-saved examples have been replayed.

### DIFF
--- a/src/hypofuzz/corpus.py
+++ b/src/hypofuzz/corpus.py
@@ -135,6 +135,18 @@ class Pool:
     def _fuzz_key(self) -> bytes:
         return self._key + b".fuzz"
 
+    @property
+    def singletons(self) -> int:
+        # Because _every_ arc hit is counted at least once, singletons are those arcs that have that base hit,
+        # and then _one_ more on top of it.
+        singletons = [item for item in self.overall_arc_counts.values() if 2 == item]
+        return len(singletons)
+
+    @property
+    def doubletons(self) -> int:
+        doubletons = [item for item in self.overall_arc_counts.values() if 3 == item]
+        return len(doubletons)
+
     def add(self, result: ConjectureResult, source: HowGenerated) -> Optional[bool]:
         """Update the corpus with the result of running a test.
 
--- a/src/hypofuzz/corpus.py
+++ b/src/hypofuzz/corpus.py
@@ -86,6 +86,8 @@ class Pool:
         self.covering_buffers: dict[Arc, bytes] = {}
         # How many times have we seen each arc since discovering our latest arc?
         self.arc_counts: Counter[Arc] = Counter()
+        # How many times have we seen each arc since start of run?
+        self.overall_arc_counts: Counter[Arc] = Counter()
 
         # And various internal attributes and metadata
         self.interesting_examples: dict[
@@ -214,11 +216,13 @@ class Pool:
         # have a different distribution with a new seed pool.
         if branches.issubset(self.arc_counts):
             self.arc_counts.update(branches)
+            self.overall_arc_counts.update(branches)
         else:
             # Reset our seen arc counts.  This is essential because changing our
             # seed pool alters the probability of seeing each arc in future.
             # For details see AFL-fast, esp. the markov-chain trick.
             self.arc_counts = Counter(branches.union(self.arc_counts))
+            self.overall_arc_counts.update(branches)
 
             # Save this buffer as our minimal-known covering example for each new arc.
             if result.buffer not in self.results:
--- a/src/hypofuzz/dashboard.py
+++ b/src/hypofuzz/dashboard.py
@@ -24,7 +24,7 @@ LAST_UPDATE: dict = {}
 
 PYTEST_ARGS = None
 
-headings = ["nodeid", "elapsed time", "ninputs", "since new cov", "branches", "note"]
+headings = ["nodeid", "elapsed time", "ninputs", "since new cov", "branches", "est. branches", "note"]
 app = flask.Flask(__name__, static_folder=os.path.abspath("pycrunch-recordings"))
 
 try:
--- a/src/hypofuzz/hy.py
+++ b/src/hypofuzz/hy.py
@@ -379,6 +379,7 @@ class FuzzProcess:
             "worker": where_am_i(),
             "ninputs": self.ninputs,
             "branches": len(self.pool.arc_counts),
+            "est. branches": "",
             "since new cov": self.since_new_cov,
             "loaded_from_db": len(self.pool._loaded_from_database),
             "status_counts": dict(self.status_counts),
--- a/src/hypofuzz/hy.py
+++ b/src/hypofuzz/hy.py
@@ -399,6 +399,13 @@ class FuzzProcess:
                 ls for _, ls in self.pool.interesting_examples.values()
             ]
             del report["since new cov"]
+
+        if self.ninputs >= 10 and "" == report["note"]:
+            singletons = self.pool.singletons
+            doubletons = self.pool.doubletons
+            offset = singletons * (singletons-1) / 2 if 0 == doubletons else singletons * singletons / (2 * doubletons)
+            offset = int(offset) + 1
+            report["est. branches"] = report["branches"] + offset
         return report
 
     @property


src/hypofuzz/corpus.py
```diff
@@ -133,6 +135,18 @@ def _check_invariants(self) -> None:
     def _fuzz_key(self) -> bytes:
         return self._key + b".fuzz"
 
+    @property
+    def singletons(self) -> int:
+        # Because _every_ arc hit is counted at least once, singletons are those arcs that have that base hit,
+        # and then _one_ more on top of it.
+        singletons = [item for item in self.overall_arc_counts.values() if 2 == item]
```
tybug, line 142: Is this true? I believe we only store arcs that we have executed (via `CustomCollectionContext`), which means singletons should be those with a count of 1, right?

tybug (commented):
Thanks @CyberiaResurrection! Could you also [sign the CLA](https://github.com/Zac-HD/hypofuzz/blob/master/CONTRIBUTING.md#contributor-license-agreement) by adding your name?

I started to make some cleanup changes to this branch, before realizing I'm not actually a maintainer. @Zac-HD do you think I could get perms? 😄

src/hypofuzz/corpus.py
CyberiaResurrection: Arcs outside of the package, like in python itself, are also counted, which (since they're only run once during a given run) will only ever have a count of 1.

CyberiaResurrection:
> Thanks @CyberiaResurrection! Could you also [sign the CLA](https://github.com/Zac-HD/hypofuzz/blob/master/CONTRIBUTING.md#contributor-license-agreement) by adding your name?
> 
Think I've now done that in "Update CONTRIBUTING.md".  Anything else you guys need?

Zac-HD:
I wrote up a longer note in https://github.com/Zac-HD/hypofuzz/issues/11#issuecomment-2558903067, but basically I'm not convinced that we can estimate reachable coverage accurately enough to be useful.  Papers on this tend to _start_ thousands of inputs into the run, and still see +100% / -50% variation against ground truth.

I think rather than estimators, I'd rather aim for _observability_ that makes it easy for users to see what's currently covered vs what isn't; e.g. though [a highlight-and-number lines view](https://www.fuzzingbook.org/html/FuzzingInTheLarge.html#:~:text=Lines%20highlighted%20in%20green%20have%20been%20executed%3B%20the%20number%20in%20the%20green%20bar%20on%20the%20left%20tells%20how%20many%20times), and then let the graph of (improved!) coverage vs log-time speak for itself.

Zac-HD:
(and, uh... it really sucks to have done a bunch of work and then have me pop in and argue against including it.  I'm sorry it took me so long to reply to the conversation; I really appreciate your engagement and contributions, and hope this isn't offputting 😭)

CyberiaResurrection:
That is a bit off-putting, @Zac-HD .  Could some of my work in this PR be salvaged?  (eg the singleton/doubleton counts)

CyberiaResurrection:
> I wrote up a longer note in [#11 (comment)](https://github.com/Zac-HD/hypofuzz/issues/11#issuecomment-2558903067), but basically I'm not convinced that we can estimate reachable coverage accurately enough to be useful. Papers on this tend to _start_ thousands of inputs into the run, and still see +100% / -50% variation against ground truth.
> 

How about only reporting estimated coverage when (say), 10k inputs into a run?  With one instance I've got locally, I'm getting 20k inputs into a run on 24.9.1 before 3 minutes are up.

tybug:
sorry to drop this conversation as well; zac knows this area better than I do and so I'll follow his recommendation here. I do think we should focus on better tests and heuristics for coverage exclusion soon, as I saw some somewhat worrying one-off compiled lines getting classified as singletons sometimes-but-not-always when testing this locally (which is not the fault of this pull!).

Zac-HD:
Yeah, memorization will create a lot of "fake singletons" if handled naively; we probably need to do some coverage-stability stuff and only consider n-tons in terms of the number of distinct inputs that trigger them (rather than raw observation count) and maybe also only for stable coverage.  See #5 for brief notes on that.

CyberiaResurrection:
@Zac-HD , I'll add tripleton and quadrupleton counts (since they were used in at least one measure in at least one of Boehme's papers), then change the bound from Chao to (second-order?) jack-knife.

I at the moment have _no_ idea how to tweak the docs to account for this change.

CyberiaResurrection:
> <snip>  as I saw some somewhat worrying one-off compiled lines getting classified as singletons sometimes-but-not-always when testing this locally (which is not the fault of this pull!).

@tybug , this is _with_ singletons requiring two hits, as per the property definition?

tybug:
That's with singletons defined as one execution. Are singletons defined as two executions in this pull to avoid this issue? I'd prefer to improve the underlying coverage counts first if so (with e.g. coverage blacklists or zac's suggestions) before reporting a metric which is actually based on n+1-tons and may be quite off as a result.

Zac-HD:
Ah, I think we have a confusion of definitions here.  If I'm right, I think

- @CyberiaResurrection is defining a singleton as "a branch which has been observed in exactly two executions" (to avoid spurious cases where we're e.g. populating a cache
- I think we need to define a singleton as "a branch which is reproducibly executed by exactly one input"

So... we need to track some notion of "which inputs have executed this branch", at least for branches where that number is small, so that we can count the singletons!  It's probably OK to revert back to counting number of times executed once we've seen >5 distinct inputs hit the branch; that's much cheaper and at that point we can assume some diversity I think.

- currently
  - on any execution where we observe a previously-unseen branch, add it to the seed pool.  Assume that replaying inputs will always have the same coverage, which is false (most often due to populating and then hitting a cache)
- alternative
  - "forkserver trick" - execute each iteration in a fresh propcess.  We'd never have cache hits from previous inputs (for better and worse), but it's tricky to set up + has performance implications + doesn't really work on Windows
- my proposal
  - when we observe a new branch, don't add that input to the pool immediately - first try to replay it.  See also my comments on #5.
    - if we get the same coverage the second time, add it to the pool.
    - otherwise, replay again - if the second to fourth are consistent and also new, add it to the pool for _that_ rather than the first coverage.
    - if it's just inconsistent, add it to a new flaky-coverage-pool; it's still worth exploring sometimes I guess even though it's unclear how.
  - because we're working off _number of distinct imputs_ now, we need to know _what inputs led to this branch_.  `.arc_counts` must therefore track `set[hash-of[choice-sequence]]`.  For efficiency, I'd switch to `| int` once the set reaches five or maybe ten elements - this is going to be annoying to deal with, so maybe we make a custom class with nice accessors?
    - A singleton is then a branch with exactly one hash in the set, doubleton has two, etc.
    - Using uint32 hash digests so that we can afford to persist this between restarts etc.  That's mostly useful when we're checking whether something changed between runs.

The downside is basically that this is a bunch more work, but I think that's what it takes to get a useful estimator (and incidentally some progress towards #5 :pray:).

CyberiaResurrection:
@Zac-HD  - is reporting the 2nd-order jack-knife estimator as in this PR, after 10k inputs, a worthwhile, if small, intermediate step towards better estimators?

Are you wanting to track inputs triggering specific branches because (iirc) that's how AFL et al do it?

Zac-HD:
> @Zac-HD - is reporting the 2nd-order jack-knife estimator as in this PR, after 10k inputs, a worthwhile, if small, intermediate step towards better estimators?

Yes!  I'm keen to get the estimator in.  I _also_ think it's important to feed it the correct observational data - GIGO, as the saying goes.

> Are you wanting to track inputs triggering specific branches because (iirc) that's how AFL et al do it?

More "I want to do it the way that works, and AFL et al happen to do the same thing because it works".  Imagine for example that Hypofuzz immediately replayed each input which reached new coverage (we don't currently do this, but we might randomly replay some old inputs soon).  If we're using execution count to define singletons _then there won't be any!_  

So we'd better switch to tracking the _number of distinct inputs_; we can safely assume that the inputs we're predicting will be unique (not quite true but close enough, and we can multiply by the duplication-rate after estimating if we want a correction), but we _can't_ assume that past inputs are all unique because Hypothesis will moderately-often replay past inputs for various reasons including those in #5.

src/hypofuzz/corpus.py
Zac-HD: We should still use `== 1` for singletons, and deal with "branches run only once" [as I describe here](https://github.com/Zac-HD/hypofuzz/pull/44#issuecomment-2585268801) by only counting branches which we can hit again by replaying the same input.

tybug:
`choices_key([1, True, ""])` (added/renamed in https://github.com/HypothesisWorks/hypothesis/pull/4241) is intended for this purpose - and I am now realizing that it does *not* distinguish 0/1 and False/True, so I'll fix that in that pull by adding an extra key to bools.

PR #46, opened 2024-12-30
Respect database in dashboard (and add tests)

tybug:
Contains a very rough start on a dashboard api - just enough to write a few basic tests.

### DIFF
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -2,7 +2,7 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
-## 24.12.1
+## 25.01.1
 
 The dashboard now respects the current setting profile's database when loading fuzzing progress.
 Use `settings.register_profile` and `settings.load_profile` to control the dashboard database.
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -2,6 +2,13 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 24.12.1
+
+The dashboard now respects the current setting profile's database when loading fuzzing progress.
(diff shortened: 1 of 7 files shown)


docs-src/changelog.md
```diff
@@ -2,6 +2,13 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 24.12.1
```
Zac-HD, line 5: 😅

src/hypofuzz/dashboard.py
```diff
@@ -307,6 +306,16 @@ def download_patch(name: str) -> flask.Response:
     )
 
 
+# api design is in flux and will likely evolve alongside our testing needs.
+@app.route("/api/state")  # type: ignore
+def api_state() -> flask.Response:
+    # todo: poll on a fixed interval (even if the page isn't loaded) instead of
+    # on every api request.
+    poll_database()
```
Zac-HD, line 314: relatively easy way to do this: have `poll_database()` record when it last polled, and be a noop if that's within a few seconds - a kinda request-coalescing-flavored TTL setup.

(longer term, we'll want to extend our database interface to support incremental queries, but that's a whole nother design question)

Zac-HD (approved):
Looks great, thanks Liam!

Zac-HD:
Also _wow_, synchronous web frameworks really draw my attention to latency-related perf issues these days.  Eventually I'll want to switch us over to something like fastapi/starlette on trio/anyio, but that's for when or maybe after we get a custom non-plotly frontend set up.  (Claude will help, I expect)

tybug:
sounds good, thouhg I can live with the current dashboard for a bit longer while we nail down more foundational issues

docs-src/changelog.md
tybug: I'll assume this is a poke to redate before merging! (and eventually set up some CI to do this for us)

PR #47, opened 2025-01-07
Use `sys.monitoring` and optimize db usage

tybug:
For the database, it's mostly `seed_pool` which can be very large, and we only need that attribute for the latest entry. (now that I write this,  we may want to stick attrs like this into a separate key `.metadata-latest` instead so we don't have to do the reduce-old-reports song and dance).

I'd like to write lots of tests for coverage - eventually. Needs its own entire pull. But if you see a missing cov instrumention test which blocks this pull, let me know.

### DIFF
--- a/src/hypofuzz/cov.py
+++ b/src/hypofuzz/cov.py
@@ -66,7 +66,7 @@ is_hypofuzz_file = belongs_to(hypofuzz)
 
 
 @cache
-def should_trace(fname):
+def should_trace(fname: str) -> bool:
     return not (is_hypothesis_file(fname) or is_hypofuzz_file(fname))
 
 
--- a/src/hypofuzz/cov.py
+++ b/src/hypofuzz/cov.py
@@ -1,11 +1,16 @@
 """Adaptive fuzzing for property-based tests using Hypothesis."""
 
 import sys
+import types
+from functools import cache
 from typing import Any, Optional
 
 import attr
 import coverage
-from hypothesis.internal.escalation import is_hypothesis_file
+import hypothesis
+from hypothesis.internal.escalation import belongs_to
+
+import hypofuzz
 
 # The upstream notion of an arc is (int, int) with an implicit filename,
 # but HypoFuzz uses an explicit filename as part of the arc.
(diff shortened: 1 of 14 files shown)


src/hypofuzz/cov.py
```diff
@@ -56,74 +61,74 @@ def get_possible_branches(cov: coverage.CoverageData, fname: str) -> frozenset[A
...
-    """Collect coverage data as a context manager.
+is_hypothesis_file = belongs_to(hypothesis)
+is_hypofuzz_file = belongs_to(hypofuzz)
 
-    The context manager can be reused; each use updates the ``.branches``
-    attribute which will be reset on next use.
 
-    TODO: excluding Hypothesis (and fuzz) files from tracing as well
-            as results would be a small performance upgrade.
-    """
```
tybug, line 67: I assume this is leftover from a coveragepy integration - hope you agree that rolling our own coverage collection instead is the way to go?

Zac-HD: yes and yes!

src/hypofuzz/hy.py
```diff
@@ -349,15 +349,34 @@ def _report(self, report: Report) -> None:
...
-            and self._last_report["branches"] == report["branches"]
-            and self._last_report["note"] == report["note"]
-            and not self.pool.interesting_examples
+        # To avoid large db sizes, we only store some fields for the latest report.
+        # Additionally, if the previous report doesn't add anything relative to
+        # this one, we drop the previous report entirely.
+        #
+        # We first unconditionally drop the previous report. If we determine we
+        # should keep it, we then re-add its reduced version, with the fields we
+        # only keep for the latest report removed.
```
Zac-HD, line 358: Reading "first" here made me think it came before writing the latest report (above), so I'd probably rewrite this.  e.g.

> Having written the latest report, we can avoid bloating the database by dropping the previous report, and re-adding a trimmed version if it differs from the latest in more than just runtime.

Zac-HD (approved):
LGTM as an improvement on the status quo, though we obviously have a lot further to go on _many_ things here :rocket:

src/hypofuzz/cov.py
```diff
@@ -56,74 +61,74 @@ def get_possible_branches(cov: coverage.CoverageData, fname: str) -> frozenset[A
...
     attribute which will be reset on next use.
-
-    TODO: excluding Hypothesis (and fuzz) files from tracing as well
-            as results would be a small performance upgrade.
     """
 
+    # tool_id = 1 is designated for coverage, but we intentionally choose a
+    # non-reserved tool id so we can co-exist with coverage tools.
+    tool_id: int = 3
+    tool_name: str = "hypofuzz"
```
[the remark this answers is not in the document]
tybug: I think it *probably* makes sense to unify these eventually, but I'm keeping them separate for now to give us freedom to experiment before settling on an interface

tybug:
I've bumped our hypothesis requirement to pull in the race condition fix - do you think that's too aggressive? My thinking is that it's not the worst thing to bump for a bugfix given that we bump compatibility with internals somewhat regularly anyway.

Zac-HD:
Yep, HypoFuzz is still in "Very Alpha" status, requiring very recent dependencies is fine.

Issue #38, opened 2024-04-25
RuleBasedStateMachine tests do not work and are not documented


pspacek:
It seems that tests based on RuleBasedStateMachine are not supported. I've tried to full-text search for "RuleBased" and "stateful" in hypofuzz docs but have not seen any mention this is not supported.

After some digging I've found a mention in the sources:
```python
# Skip state-machine classes, since they're not
```
but it's a bit mysterious :-)

Maybe collector could print a message when in encounters unsupported case instead of silently ignoring it?

### Versions tested
- hypofuzz 24.2.3
- hypothesis 6.100.1
- pytest 8.0.2

### Steps to reproduce
1. Copy & paste example code from https://hypothesis.readthedocs.io/en/hypothesis-python-4.57.1/stateful.html into a file, say `test_example.py`.

2. Run `hypothesis fuzz test_example.py`

### Output
```
Usage: hypothesis fuzz [OPTIONS] [-- PYTEST_ARGS]
Try 'hypothesis fuzz -h' for help.

Error: No property-based tests were collected
```

### Further details
With a slight modification to `hypofuzz/interface.py` we can get more verbose output. Diff:
```diff
@@ -81,8 +81,8 @@ def _get_hypothesis_tests_with_pytest(args: Iterable[str]) -> List["FuzzProcess"
             ],
             plugins=[collector],
         )
+    print(out.getvalue())  # noqa
     if ret:
-        print(out.getvalue())  # noqa
         print(f"Exiting because pytest returned exit code {ret}")  # noqa
         sys.exit(ret)
     return collector.fuzz_targets
```

Output:
```
================================ test session starts =================================
platform linux -- Python 3.11.8, pytest-8.0.2, pluggy-1.5.0
rootdir: /tmp/ste/orig
plugins: hypothesis-6.100.1, dash-2.16.1
collected 1 item

<Dir orig>
  <Module test_example.py>
    <UnitTestCase TestDBComparison>
      <TestCaseFunction runTest>
crashed in test_example.py::TestDBComparison::runTest 'function' object has no attribute 'hypothesis'

============================= 1 test collected in 0.02s ==============================
```

pspacek:
Additional note, after commenting out `print` around catch-all `except` I get more detailed traceback:
```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/tmp/ste/hypofuzz/lib/python3.11/site-packages/_pytest/main.py", line 273, in wrap_session
[...]

PR #48, opened 2025-01-20
Support autouse fixtures that request another fixture

tybug:
This happens when fuzzing Hypothesis itself, as our `_consistently_increment_time` autouse fixture uses the `monkeypatch` fixture. See also https://github.com/Zac-HD/hypofuzz/issues/38#issuecomment-2188172857.

### DIFF
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -2,6 +2,10 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 25.01.3
+
+Support collecting tests with a pytest autouse fixture that itself requests a non-autouse fixture.
+
 ## 25.01.2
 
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
Gah, we're going to have to completely support fixtures at some point, and I'm _really_ not looking forward to that.

Thanks for the patch.

tybug:
any high level idea of how we're going to do that? Will we have to replicate pytest's lifecycle (cleanup/teardown) ourselves? I started a bit on supporting function-scoped fixtures, which should get us 90% of the way there...

Zac-HD:
yeah, I think it's basically replicating the relevant parts of Pytest.  If we think of a nice clean abstraction that would make it easier I'd be happy to sponsor that upstream, but I suspect that it's just going to be a slog.

Upside, if we get function-scoped fixtures working in Hypothesis we can probably reuse that for all HypoFuzz purposes 🤷

PR #49, opened 2025-01-21
Ignore stdlib and dynamic code

tybug:
I know there's some debate about using stdlib code as fuzzing guidance. (un)fortunately for us, the point is rather moot, because *hypofuzz itself* (and its dependencies, like hypothesis) uses the stdlib in its implementation, which we don't want to trace. Rather than use some complicated whitelist or tracing callstacks, let's just disable it entirely.


Zac-HD (approved):
makes sense!

PR #50, opened 2025-01-23
Initial support for parameterization and skipif

tybug:
Parametrized node ids Just Work ™️ in the dashboard....as long as you aren't doing anything too cursed, like duplicating names. Skipif is evaluated and skipped during collection, including support for condition strings.

I haven't tested this with more complicated things like classes or ~~module-level skips~~ (added tests for the latter). 

Also tacked on to this pull: update our database usage from buffer to choices.

### DIFF
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -2,6 +2,10 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 25.01.5
+
+Initial support for `@pytest.mark.parametrize` and `@pytest.mark.skipif`.
+
 ## 25.01.4
 
 Ignore standard library and dynamically generated code for coverage.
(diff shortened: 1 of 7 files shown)


Zac-HD (approved):
Looks great - thanks Liam!  I'm pretty happy to support only Pytest 7+, which we're already testing on, so I think this is good to go 🎉

tybug:
ah, I actually didn't realize something in this pull broke support for pytest 6? Do you think we're no longer compatible with pytest 6? If so I'll update our `setup.py` requirements to 7.0.0

Zac-HD:
Based on [download stats](https://pepy.tech/projects/Pytest?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=weekly&viewType=chart&versions=6.*%2C7.*%2C8.*), I'm ok with hypofuzz supporting pytest 7+ only, at least for now.  I don't know if older versions still work, but we should probably declare what we test in CI as the minimum bound...

PR #51, opened 2025-01-23
Initial support for stateful testing

### LINKED ISSUE #38: RuleBasedStateMachine tests do not work and are not documented

tybug:
Closes #38. Depends on #50 for a clean merge.

This requires a small amount of refactoring Hypothesis-side, including setting a new `runTest._hypothesis_state_machine_class = cls` attribute. I don't have a *great* grasp of the stateful testing internals, so I may be overcomplicating things? It seems that what Hypofuzz really wants to access is the underlying `@given` which drives stateful test - and to get to that we need to rejig some code.

Here's the Hypothesis-side companion diff, which I can PR to Hypothesis:

```diff
diff --git a/hypothesis-python/src/hypothesis/stateful.py b/hypothesis-python/src/hypothesis/stateful.py
index f1dfa5205..ae8ccd36c 100644
--- a/hypothesis-python/src/hypothesis/stateful.py
+++ b/hypothesis-python/src/hypothesis/stateful.py
@@ -83,14 +83,7 @@ class TestCaseProperty:  # pragma: no cover
         raise AttributeError("Cannot delete TestCase")
 
 
-def run_state_machine_as_test(state_machine_factory, *, settings=None, _min_steps=0):
-    """Run a state machine definition as a test, either silently doing nothing
-    or printing a minimal breaking program and raising an exception.
-
-    state_machine_factory is anything which returns an instance of
-    RuleBasedStateMachine when called with no arguments - it can be a class or a
-    function. settings will be used to control the execution of the test.
-    """
+def get_state_machine_test(state_machine_factory, *, settings=None, _min_steps=0):
     if settings is None:
         try:
             settings = state_machine_factory.TestCase.settings
@@ -237,8 +230,21 @@ def run_state_machine_as_test(state_machine_factory, *, settings=None, _min_step
         state_machine_factory, "_hypothesis_internal_use_reproduce_failure", None
     )
     run_state_machine._hypothesis_internal_print_given_args = False
+    return run_state_machine
+
 
-    run_state_machine(state_machine_factory)
+def run_state_machine_as_test(state_machine_factory, *, settings=None, _min_steps=0):
+    """Run a state machine definition as a test, either silently doing nothing
+    or printing a minimal breaking program and raising an exception.
+
+    state_machine_factory is anything which returns an instance of
+    RuleBasedStateMachine when called with no arguments - it can be a class or a
+    function. settings will be used to control the execution of the test.
+    """
+    state_machine_test = get_state_machine_test(
+        state_machine_factory, settings=settings, _min_steps=_min_steps
+    )
+    state_machine_test(state_machine_factory)
 
 
 class StateMachineMeta(type):
@@ -437,6 +443,7 @@ class RuleBasedStateMachine(metaclass=StateMachineMeta):
                 run_state_machine_as_test(cls, settings=self.settings)
 
             runTest.is_hypothesis_test = True
+            runTest._hypothesis_state_machine_class = cls
 
         StateMachineTestCase.__name__ = cls.__name__ + ".TestCase"
         StateMachineTestCase.__qualname__ = cls.__qualname__ + ".TestCase"
```

### DIFF
--- a/src/hypofuzz/interface.py
+++ b/src/hypofuzz/interface.py
@@ -4,14 +4,13 @@ import io
 import sys
 from collections.abc import Iterable
 from contextlib import redirect_stdout
-from functools import partial
 from inspect import signature
 from typing import TYPE_CHECKING, get_type_hints
 
 import pytest
 from _pytest.nodes import Node
 from _pytest.skipping import evaluate_condition
-from hypothesis.stateful import RuleBasedStateMachine, run_state_machine_as_test
+from hypothesis.stateful import get_state_machine_test
 
 if TYPE_CHECKING:
     # We have to defer imports to within functions here, because this module
@@ -71,6 +70,7 @@ class _ItemsCollector:
             # values directly, so we can pass them as extra kwargs to FuzzProcess.
             params = item.callspec.params if hasattr(item, "callspec") else {}
             param_names = set(params)
+            extra_kw = params
 
             # Skip any test which:
             # - directly requests a non autouse fixture, or
@@ -95,16 +95,23 @@ class _ItemsCollector:
                     flush=True,
                 )
                 continue
-
             # Wrap it up in a FuzzTarget and we're done!
             try:
-                # Skip state-machine classes, since they're not
-                if isinstance(item.obj, RuleBasedStateMachine.TestCase):
-                    target = partial(run_state_machine_as_test, item.obj)
+                if hasattr(item.obj, "_hypothesis_state_machine_class"):
+                    # I guess it's conceivable this is possible...I just don't see how!
+                    assert (
+                        extra_kw == {}
+                    ), "Not possible for RuleBasedStateMachine.TestCase to be parametrized (we think?)"
+                    runTest = item.obj
+                    StateMachineClass = runTest._hypothesis_state_machine_class
+                    target = get_state_machine_test(
+                        StateMachineClass, settings=runTest.__self__.settings
+                    )
+                    extra_kw = {"factory": StateMachineClass}
                 else:
                     target = item.obj
                 fuzz = FuzzProcess.from_hypothesis_test(
-                    target, nodeid=item.nodeid, extra_kw=params
+                    target, nodeid=item.nodeid, extra_kw=extra_kw
                 )
                 self.fuzz_targets.append(fuzz)
             except Exception as err:
--- a/src/hypofuzz/interface.py
+++ b/src/hypofuzz/interface.py
@@ -11,11 +11,14 @@ import pytest
 from _pytest.nodes import Item, Node
 from _pytest.skipping import evaluate_condition
 from hypothesis.stateful import get_state_machine_test
+from packaging import version
 
 if TYPE_CHECKING:
     # We have to defer imports to within functions here, because this module
     # is a Hypothesis entry point and is thus imported earlier than the others.
     from .hy import FuzzProcess
+
+pytest8 = version.parse(pytest.__version__) >= version.parse("8.0.0")
 
 
 def has_true_skipif(item: Item) -> bool:
@@ -110,7 +113,17 @@ class _ItemsCollector:
                     runTest = item.obj
                     StateMachineClass = runTest._hypothesis_state_machine_class
                     target = get_state_machine_test(  # type: ignore
-                        StateMachineClass, settings=runTest.__self__.settings
+                        StateMachineClass,
+                        # runTest is a function, not a bound method, under pyest7.
+                        # I wonder if something about TestCase instantiation order
+                        # changed in pytest 8? Either way, we can't access
+                        # __self__.settings uder pytest 7.
+                        #
+                        # I am going to call this an acceptably rare bug for now,
+                        # because it should only manifest if the user sets a custom
+                        # database on a stateful test under pytest 7 (all non-db
+                        # settings are ignored by hypofuzz).
+                        settings=runTest.__self__.settings if pytest8 else None,
                     )
                     extra_kw = {"factory": StateMachineClass}
                 else:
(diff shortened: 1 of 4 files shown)


src/hypofuzz/interface.py
```diff
@@ -95,16 +95,23 @@ def pytest_collection_finish(self, session: pytest.Session) -> None:
...
             # Wrap it up in a FuzzTarget and we're done!
             try:
-                # Skip state-machine classes, since they're not
-                if isinstance(item.obj, RuleBasedStateMachine.TestCase):
-                    target = partial(run_state_machine_as_test, item.obj)
+                if hasattr(item.obj, "_hypothesis_state_machine_class"):
+                    # I guess it's conceivable this is possible...I just don't see how!
+                    assert (
+                        extra_kw == {}
+                    ), "Not possible for RuleBasedStateMachine.TestCase to be parametrized (we think?)"
```
Zac-HD, line 104: Correct, because Hypothesis is responsible for creating the `TestCase` object, and it's not collected by pytest iirc.

src/hypofuzz/interface.py
```diff
@@ -95,16 +95,23 @@ def pytest_collection_finish(self, session: pytest.Session) -> None:
                     flush=True,
                 )
                 continue
-
             # Wrap it up in a FuzzTarget and we're done!
             try:
-                # Skip state-machine classes, since they're not
-                if isinstance(item.obj, RuleBasedStateMachine.TestCase):
-                    target = partial(run_state_machine_as_test, item.obj)
+                if hasattr(item.obj, "_hypothesis_state_machine_class"):
```
Zac-HD, line 100: Rather than attaching a magic attribute - which could be confused by mocks or whatever - let's define a subclass of `TestCase` in `hypothesis.stateful` and then [inherit from it here](https://github.com/HypothesisWorks/hypothesis/blob/eaa0852472ff4830782bfabb9043b914ef989332/hypothesis-python/src/hypothesis/stateful.py#L428-L443), so we have a principled way to identify it on the Hypofuzz side.

Zac-HD:
This will be very nice to have, and the Hypothesis-side refactoring looks good to me too (with one note above).

tybug:
will look to merge this after #53; the timing of the hypothesis-side releases means we need compatibility with bufferless hypothesis first

tybug:
Error only in pytest 7 🤔 will have to figure out what internals changed here

```
crashed in test_wl561fdj.py::NumberModifierTest::runTest 'function' object has no attribute '__self__'
```

tybug:
I gave up because 1) this seems like it could be a rabbit hole to track down, and 2) the requirements to trigger this are rare (pytest 7 + stateful + custom database)

src/hypofuzz/interface.py
```diff
@@ -101,16 +104,32 @@ def pytest_collection_finish(self, session: pytest.Session) -> None:
...
+                        # runTest is a function, not a bound method, under pyest7.
+                        # I wonder if something about TestCase instantiation order
+                        # changed in pytest 8? Either way, we can't access
+                        # __self__.settings uder pytest 7.
+                        #
+                        # I am going to call this an acceptably rare bug for now,
+                        # because it should only manifest if the user sets a custom
+                        # database on a stateful test under pytest 7 (all non-db
+                        # settings are ignored by hypofuzz).
+                        settings=runTest.__self__.settings if pytest8 else None,
```
Zac-HD, line 126: Ugh, I really don't like this but I think it's probably the best option we practically have.

In the medium term, I'd like us to track all the tests-we-don't-collect, and have a dashboard page which lists them and explains why each was skipped (as well as optional terminal reporting?  hmm).

Zac-HD (approved):

src/hypofuzz/interface.py
tybug: yup I'd like that as well, should be doable nearish term. (our dashboard will probably turn into the standard "sidebar on the left which lists pages", and one page is a test overview or similar).

I will come back to this bug at some point, there are just many better things to do in the meantime!

tybug: I think I found the relevant pytest change: https://github.com/pytest-dev/pytest/pull/12089. unsure if a fix will be easy

Zac-HD: let's make sure we pick frameworks etc which can be responsive / mobile-friendly; the dominant usecase will be on localhost but scaling beyond that is a goal.

Issue #36, opened 2024-04-22
Split the HypoFuzz engine into a Hypothesis `backend` and an executor 


Zac-HD:
Hypothesis has recently grown a notion of [alternative backends](https://hypothesis.readthedocs.io/en/latest/strategies.html#alternative-backends), which use the new IR layer (https://github.com/HypothesisWorks/hypothesis/issues/3921) to support e.g. symbolic execution (https://github.com/HypothesisWorks/hypothesis/issues/3914).

Supporting `@settings(backend="hypofuzz")` would be quite useful - for example, as an easy way to work with Pytest fixtures, or to support fuzzing in environments where Pytest is not available at all (CPython alphas?).

The implementation isn't trivial, since we have to rearrange a lot of code to both fit with the IR-oriented, context-manager-scoped interface for backends - while also having a hook to allow for interleaving execution of multiple tests - but I don't think there are any fundamental difficulties.

This change is independent of #3, but strongly complimentary in practice.

PR #53, opened 2025-02-02
Move to a typed choice sequence mutator

tybug:
Ditches the buffer mutator in favor of an (extremely simple) tcs mutator.

TBH I have relatively low confidence that the logic in `corpus.py` is correct after this change. I see more test writing in my future...

### DIFF
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -6,7 +6,7 @@ HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 
 Use a new mutator based on the typed choice sequence (https://github.com/HypothesisWorks/hypothesis/issues/3921), bringing back compatibility with new Hypothesis versions.
 
-We also now ignore the `pytest` and `_pytest` during coverage collection.
+We also now ignore code coverage inside the `pytest` and `_pytest` modules, which is unhelpful to fuzz.
 
 ## 25.01.6
 
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -6,6 +6,8 @@ HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 
 Use a new mutator based on the typed choice sequence (https://github.com/HypothesisWorks/hypothesis/issues/3921), bringing back compatibility with new Hypothesis versions.
 
+We also now ignore the `pytest` and `_pytest` during coverage collection.
+
 ## 25.01.6
 
 Pin upper Hypothesis bound for compatibility, until we support new Hypothesis internals.
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -2,6 +2,10 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 25.02.1
+
+Use a new mutator based on the typed choice sequence (https://github.com/HypothesisWorks/hypothesis/issues/3921), bringing back compatibility with new Hypothesis versions.
+
 ## 25.01.6
 
 Pin upper Hypothesis bound for compatibility, until we support new Hypothesis internals.
--- a/src/hypofuzz/corpus.py
+++ b/src/hypofuzz/corpus.py
@@ -3,7 +3,7 @@
 import abc
 import enum
 from collections import Counter
-from collections.abc import Callable, Iterable
+from collections.abc import Callable, Iterable, Iterator
 from random import Random
 from typing import TYPE_CHECKING, Optional, Union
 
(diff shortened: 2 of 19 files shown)


tybug:
yeah in fact adding `hypothesis` to the `tox -e check` job dependencies turns up several bugs in my implementation - mostly mixing args of types `IRNode` and `ChoiceT` in `cov.py`. I will add some real tests here before considering merging

tybug:
OK, I'm much more confident in these changes now that we have some tests. Still have some type hints to take care of.

docs-src/changelog.md
```diff
@@ -2,6 +2,12 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 25.02.1
+
+Use a new mutator based on the typed choice sequence (https://github.com/HypothesisWorks/hypothesis/issues/3921), bringing back compatibility with new Hypothesis versions.
+
+We also now ignore the `pytest` and `_pytest` during coverage collection.
```
Zac-HD, line 9:
```suggestion
We also now ignore code coverage inside the `pytest` and `_pytest` modules, which is unhelpful to fuzz.
```

src/hypofuzz/corpus.py
```diff
@@ -3,53 +3,93 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Choices):
+            return NotImplemented
+        return choices_key(self.choices) == choices_key(other.choices)
+
+    def __str__(self) -> str:
+        return str(self.choices)
+
+    __repr__ = __str__
```
Zac-HD, line 55: maybe `return f"Choices({self.choices!r})"`?

src/hypofuzz/cov.py
```diff
@@ -34,6 +36,11 @@ def make(fname: str, start: int, end: int) -> "Arc":
             _ARC_CACHE.setdefault(fname, {}).setdefault(start, {})[end] = self
             return self
 
+    def __str__(self) -> str:
+        return f"{self.fname}:{self.start_line}:{self.end_line}"
```
Zac-HD, line 40:
```suggestion
        return f"{self.fname}:{self.start_line}::{self.end_line}"
```
The conventional reading would be `fname:lineno:colno`, so I think we want `::` to distinguish this.  Maybe in future we'll get column number too from branch observations?  

(Avoiding redundancy for branchless code would be nice too; iirc there were some comments on Hypothesis about this.)

src/hypofuzz/cov.py
```diff
@@ -50,6 +57,8 @@ def get_coverage_instance(**kwargs: Any) -> coverage.Coverage:
 
 is_hypothesis_file = belongs_to(hypothesis)
 is_hypofuzz_file = belongs_to(hypofuzz)
+is_pytest_file = belongs_to(pytest)
+is__pytest_file = belongs_to(_pytest)
```
Zac-HD, line 61: Hmm, we should check that `belongs_to()` memoizes so that the returned functions share a cache.

src/hypofuzz/provider.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def fresh_choice(
+    choice_type: ChoiceNameT, kwargs: ChoiceKwargsT, *, random: Random
+) -> ChoiceT:
+    cd = ConjectureData(random=random)
+    return getattr(cd.provider, f"draw_{choice_type}")(**kwargs)
+
+
+class HypofuzzProvider(PrimitiveProvider):
```
Zac-HD, line 28: Just noting that https://github.com/Zac-HD/hypofuzz/issues/36 will involve going notably further, in putting much of the corpus logic etc. into the provider, and setting `lifetime = "test_function"`.  That shouldn't be in this PR, just want to avoid any confusion for future readers.

Zac-HD (commented):
🚀

src/hypofuzz/provider.py
tybug: We'll see how much we can move into the provider...Hypofuzz wants control over e.g. intercepting interesting results, and how many examples are generated, so I'm not sure we can fully move to `settings(backend="hypofuzz")(test_function)(**extra_kw)`. I think we either give up control over control flow from hypofuzz, or add provider hooks that alter control flow (`on_failing_example_found` which can raise a control flow exception "actually, I don't want you to shrink this"). I'd prefer not to do the latter; I think provider should be almost entirely read-only, not explicitly writing back to hypothesis control flow.

Fully moving to a provider could be ok if we accept the flow is "when a process fuzzes f, run a constant 1k examples and then defer back to our process pool, and any failures are shrunk immediately".

...except now that I write this, I don't know how we would run a distill phase with this proposed structure. We'd need to keep a reference to the underlying function/kwargs around, and avoiding that is kinda the point of the backend.

src/hypofuzz/cov.py
tybug: each `belongs_to` has its own cache, not shared. Would sharing help? Knowing that an `fpath: str` does or doesn't belong to `pytest` doesn't tell us anything about whether it belongs to `_pytest`, right?

Zac-HD: I mean we have `is_hypothesis_file = belongs_to(hypothesis)` defined in both Hypothesis and HypoFuzz, and iirc those are cached separately.  Not a big deal either way, of course.

Zac-HD (approved):

src/hypofuzz/cov.py
tybug: Ah, right. Yeah, we probably should make `belongs_to` share a cache, and make it lru while we're at it.

src/hypofuzz/cov.py
tybug: I'd like to compress the arcs for sure, I think we will have to ast analyze though. will probably start by relying on coverage.py for that

Issue #17, opened 2023-03-05
Incorrect data aggregation for dashboard plots?


Zac-HD:
I got a bug report by email, saying that there's a bug in the dashboard where the lengths of columns are different to the length of the index, for various columns in some pandas `DataFrame` inside the `px.line()` call. Perhaps sometimes only some keys are appended?

Putting this on a more principled (and tested!) basis might be nice even before the big changes in #3.

tybug:
Fixed by the new dashboard in #54 !

Issue #45, opened 2024-12-21
24.11.1 not clearing coverage history between restarts


CyberiaResurrection:
As can be seen in the screenshot below, on Fedora 40, hypofuzz 24.11.1 with python 3.10 and hypothesis 6.122.1 isn't clearing the coverage history on a restart, at least for the case of running exactly one hypothesis test under hypofuzz:

![image](https://github.com/user-attachments/assets/edcf0b7c-3183-4b71-ba3b-79dc9599ea96)

Each of the vertical drops downwards is another hypofuzz restart.

24.9.1 did _not_ have this problem, at least with the python and hypothesis versions listed - it cleared coverage history on each restart.

Zac-HD:
As of PR #41, that's the intended behavior!  We _want_ to persist coverage information etc. between restarts, because this makes our estimators much more accurate and thus we can allocate time between multiple tests more accurately.

On the other hand, we're well aware that this isn't the ideal way to display it on the dashboard.  There are some ideas for how we could improve this in issue #3, but basically we're bottlenecked on someone having time to experiment with analysis of the data we're now collecting, and then implementing that for the dashboard.  This is tricky because restarts are the _simplest_ case we need to handle; there's also the possibility of multiple workers running at the same time, and of the behavior of the system under test changing between runs.

-------------------

So with that context, what would you want to see on this summary page, and/or on the test-specific page?

I can imagine wanting to toggle between latest-run-only (ie old version), all-runs (ie this version), reconstructed-as-if-it-was-all-one-run (not yet implemented), better distinguishing between replaying known inputs and generating new ones...

CyberiaResurrection:
Yeah, I would like to (at least have the option to) restore the pre 24.11 behaviour (displaying latest-run-only) on both the summary and per-test page.

tybug:
I think we could implement a latest-run-only toggle more cleanly with a combination of marks on the x axis and the already existing dragclick-to-subset-graph functionality. The x axis markers (alternatively: colored shading?) could indicate the start of a new run, or phases within that run (replay, generate, distill). Clicking a marker for a run might subset the graph into it. Lots of options to explore here.

Zac-HD:
The [latest version of the dashboard](https://hypofuzz.com/example-dashboard/) is (imo) much improved; we have some pretty good logic for combining reports from multiple runs _without_ counting the "dead time" of replaying covering examples to get back up to the previous state:

[...]

PR #54, opened 2025-02-14
Move to a react dashboard

### LINKED ISSUE #7: Explore other dashboard tech options
Maybe I just need to refactor things and then Plotly Dash will be satisfyingly declarative and performant.

Alternatively I might want to explore other frameworks - e.g. Bokeh, or go up a level to Holoviews.  Also worth checking whether using Pyodide in the browser is a better user experience?  I think I'm still better off sticking to Python over JS for this, since I'm much better at the former.

tybug:
Closes #7, #17. Most of the react code was written by claude, and touched up by me.

* The stack is trio + starlette + hypercorn + react + scss
* For building/distributing: 
  * I've moved entirely to `pyproject.toml` for building, with a custom build hook, and am requiring the developer have `npm` installed in their host system to build. 
  * For distribution, I'm building the .html in `dist/*` as part of the build step and am uploading that as part of the compiled `.whl.` It should just work ™️ when installed normally (untested).
* I'm following [BEM](https://getbem.com/) for structuring scss rules. I've used this style guide before and have enjoyed it, but am not opposed to an alternative. (to be clear, this is not a running dependency or anything, just a style guide).

TODO:
- [x] check patches/pycrunch recordings, I've probably broken these
- [x] feature parity with old dashboard: per-test covering examples somewhere, `since new cov` on test overview page
- [x] unify overview / test coverage graph styling
- [x] coverage details on hover in graph

This dashboard is still missing some stuff compared to the existing one, but it is runnable 😄

image


Zac-HD:
eh, I think I'm OK with taking out `pycrunch` if it's annoying to update; it's a really nice concept but is early-alpha stability and there's no nice way to support it with distributed fuzzing (since the recordings are often hundreds of megabytes)

I'm also increasingly keen on the reconstructed-as-if-one-run view (cf https://github.com/Zac-HD/hypofuzz/issues/45#issuecomment-2558353193), but that's definitely a future PR.

tybug:
pycrunch/pytrace should be doable to update. I'll do that, and then we should ship this, since it also fixes compatibility with newer hypothesis versions (hopefully the last of those now that namings have settled down).

Zac-HD (approved):
I've spun this up locally, and it looks great!  I'll leave some more notes with followups below, but my inclination now is to merge this and take smaller incremental PRs to continue improving 😁

Zac-HD:
Random notes from playing around with this:

- obviously we'll want to use the ✨new pubsub db feature✨
- zoomable / selectable graphs are pretty important
- it'd be cool if clicking on a line on the graph took you to that test's page
- the patches page says "waiting"; something like "no failing examples yet, so no patches either" might be clearer
- in the minimal failing examples, I noticed that for tests using `st.data()` the interactive draws weren't printed
  - and as a more general principle, I think we should discard textually-identical entries in the list, since they're ~useless to the end user
- I'd love to get a live/interactive demo on the website, and iirc we chatted about doing this by having the frontend consume data from either the python backend or a static json blob (with timing metadata?).  They say you can solve (or at least move) almost any problem by adding another layer of abstraction...
- maybe we should be running ts tooling in CI?  e.g. prettier for formatting?  Plus updating pinned deps.  IDK, not an ecosystem I'm familiar with.

tybug:
lots of good stuff in that list, will work through it 👍. Brief notes:
* +1 for prettier, just wanted to leave tooling for a higher scrutiny review
* clicking on a graph line should already work (but hover tooltip ux needs work)

PR #58, opened 2025-02-20
Build before release

tybug:
I had this right the first time!


tybug:
Whew, this one worked! I declare success on building and distributing a react app as html.

PR #59, opened 2025-02-23
Use hypothesis pubsub

tybug:
* deletions with an unknown value are expensive to rescan, so I've debounced this pretty heavily at 5 minutes (and could go further). It only matters for dropping intermediary points for performance, so timing isn't too important.
* I've split database entries into `Report` and `Metadata`. The latter is where "items which are expensive and I only need to know their very latest state" live, like the seed pool. Only transmitting Reports through the websocket, not Metadata, helps a lot with bandwidth usage. Also helps db size because we were previously leaving around some `seed_pool` entries in intermediary reports if you stopped+started the server a bunch.

Depends on https://github.com/HypothesisWorks/hypothesis/pull/4270 (I'm going to look into the flaky linux failure in that pr next)

### DIFF
--- a/src/hypofuzz/dashboard.py
+++ b/src/hypofuzz/dashboard.py
@@ -1,7 +1,10 @@
 """Live web dashboard for a fuzzing run."""
 
+import abc
 import atexit
+import json
 import signal
+import time
 from collections import defaultdict
 from pathlib import Path
 from typing import Any, Optional
@@ -10,6 +13,8 @@ import black
 import trio
 from hypercorn.config import Config
 from hypercorn.trio import serve
+from hypothesis.database import ListenerEventT
+from sortedcontainers import SortedList
 from starlette.applications import Starlette
 from starlette.middleware import Middleware
 from starlette.middleware.cors import CORSMiddleware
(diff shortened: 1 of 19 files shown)


src/hypofuzz/database.py
```diff
@@ -1,23 +1,59 @@
...
+    pod_ip: Optional[str]
+    container_id: Optional[str]
 
-def metadata_key(key: bytes) -> bytes:
-    return key + b".hypofuzz.metadata"
+
+# Conceptually:
+# * A report is an incremental progress marker which we don't want to delete,
+#   because seeing intermediary stages in e.g. a graph is useful information
+# * Ketadata is the latest status of a test, which we might update to something
```
Zac-HD, line 23:
```suggestion
# * Metadata is the latest status of a test, which we might update to something
```

src/hypofuzz/dashboard.py
```diff
@@ -19,32 +24,160 @@
...
+        key, value = args
+        assert value is not None
+        if key.endswith(reports_key):
+            report = json.loads(value)
+            REPORTS[report["nodeid"]].add(report)
+            trio.run(broadcast_event, "save", report)
+        if key.endswith(metadata_key):
+            metadata = json.loads(value)
+            METADATA[metadata["nodeid"]] = metadata
+            trio.run(broadcast_event, "metadata", metadata)
```
Zac-HD, line 153: I'm pretty suspicious of these `trio.run` calls - I think it actually is safe in this case, because we're just sending some websocket events without other interaction, but it's inelegant to turn an async operation into a blocking sync operation.

Instead: we can `trio.open_memory_channel(inf)`, and `db.add_listener(send_chan.send_nowait)`; then start_soon a task in the `run_dashboard` nursery which does `async for event in recv_chan: <this handling logic>`.

src/hypofuzz/dashboard.py
```diff
@@ -166,18 +273,19 @@ def start_dashboard_process(
     # we run a pytest collection step for the dashboard to pick up on databases
     # from custom profiles.
     _get_hypothesis_tests_with_pytest(pytest_args)
+    get_db()._db.add_listener(database_listener)
 
     # Ensure that we dump whatever patches are ready before shutting down
     def signal_handler(signum, frame):  # type: ignore
-        make_and_save_patches(pytest_args, LAST_UPDATE)
+        make_and_save_patches(pytest_args, REPORTS, METADATA)
```
Zac-HD, line 280: I think with our database-centric design now there's not much point writing out patches on signal or atexit.  If you can't think of anything either, let's delete this for simplicity?

Zac-HD (commented):
We're very close now!

src/hypofuzz/dashboard.py
tybug: great point - I should have mentioned that I am not that good at async and so critiques of trio usages are very welcome. (I didn't even know open_memory_channel was a thing). I did see some trio failures if you refresh the dashboard fast enough which may well be from this call...

src/hypofuzz/dashboard.py
tybug: yeah I have "refactor patches" on my todo list which I think will involve removing this signal...we can do that now

src/hypofuzz/dashboard.py
Zac-HD: Yep, that seems pretty likely to me!  

IME Trio is good at making easy things easy, and we should be pretty solidly in the easy-things-only regime for our little webapp.  As a general tip though, idiomatic use often leans heavily on context managers to ensure cleanup, and channels to communicate between tasks or act as a buffer (where that's necessary - and mostly here it's not).  Avoid async generators; see PEP-789 for details if curious.

tybug: I'm still working through this one, but in case you have ideas - `watchdog` uses a background thread to track events, and `get_db()._db.add_listener(send_channel.send_nowait)` complains with `RuntimeError: must be called from async context` when that thread tries to send an event

Zac-HD: Try `partial(trio.from_thread.run_sync, send_chan.send_nowait, trio_token=trio.lowlevel.current_trio_token())` ([docs](https://trio.readthedocs.io/en/stable/reference-core.html#trio.from_thread.run_sync))

src/hypofuzz/dashboard.py
```diff
@@ -149,10 +216,65 @@ async def serve_app(app: Any, host: str, port: str) -> None:
...
+        # DirectoryBasedExampleDatabase sends events from a background thread (via watchdog).
+        partial(
+            trio.from_thread.run_sync,
+            send_channel.send_nowait,  # type: ignore
+            trio_token=trio.lowlevel.current_trio_token(),
+        )
+        # two layers of unwrapping: db is a HypofuzzDatabase(BackgroundWriteDatabse(<actual database here>))
+        if isinstance(db._db._db, DirectoryBasedExampleDatabase)  # type: ignore
+        else send_channel.send_nowait
+    )
```
Zac-HD, line 274: I'd be sad to break user-defined databases which run the listener in a thread but don't exactly match this pattern; fortunately I think we can handle that with a listener function that checks where it's called:
```suggestion
    token = trio.lowlevel.current_trio_token()

    def send_nowait_from_anywhere(msg: object) -> None:
        # DirectoryBasedExampleDatabase sends events from a background thread (via watchdog),
        # so we need to support sending from anywhere, i.e. whether or not the calling thread
        # has any Trio state.  We can do that with the following branch: 
        try:
            trio.lowlevel.current_task()
        except RuntimeError:
            trio.from_thread.run_sync(send_channel.send_nowait, msg, trio_token=token)
        else:
            send_channel.send_nowait(msg)

    db = get_db()
    db._db.add_listener(send_nowait_from_anywhere)
```

Zac-HD (approved):
trickier Trio interaction than I was expecting!  But I think this should work...

src/hypofuzz/dashboard.py
Zac-HD: oops, apparently I should have typed `msg: ListenerEventT` 😅

src/hypofuzz/dashboard.py
```diff
@@ -149,10 +216,68 @@ async def serve_app(app: Any, host: str, port: str) -> None:
...
+
+            # We're only keeping a reference to the latest metadata, so we don't need
+            # to bother handling deletion events for it, just save.
+
+
 async def run_dashboard(port: int, host: str) -> None:
+    send_channel, receive_channel = trio.open_memory_channel[ListenerEventT](math.inf)
+    token = trio.lowlevel.current_trio_token()
+
+    def send_nowait_from_anywhere(msg: object) -> None:
```
Zac-HD, line 265:
```suggestion
    def send_nowait_from_anywhere(msg: ListenerEventT) -> None:
```

src/hypofuzz/dashboard.py
tybug: sweet ty

PR #68, opened 2025-03-20
Add pytest collection status page

tybug:
- [x] Depends on https://github.com/Zac-HD/hypofuzz/pull/64 for ease of conflict resolution.

Adds a `/collected` page which shows collection status of all tests:

image


tests/test_collection.py
```diff
@@ -24,7 +24,9 @@ def collect(code: str) -> list[FuzzProcess]:
     )
     p = Path(tempfile.mkstemp(prefix="test_", suffix=".py")[1])
     p.write_text(code)
-    fps = interface._get_hypothesis_tests_with_pytest([str(p), "-p", "no:dash"])
+    fps = interface._get_hypothesis_tests_with_pytest(
+        [str(p), "-p", "no:dash"]
```
Zac-HD, line 28: Oh, we can drop the `-p no:dash` now; that was to disable the plotly-dash pytest plugin.  Worth searching more generally to see if we've left it anywhere else too.

src/hypofuzz/interface.py
```diff
@@ -30,24 +32,39 @@ def has_true_skipif(item: Item) -> bool:
...
-                print(f"skipping non-pytest.Function item {item=}")
+                self._skip_because("not_a_function", item.nodeid)
                 continue
             if list(item.iter_markers("skip")):
-                print(f"skipping {item=} due to an @skip mark")
+                self._skip_because("skip", item.nodeid)
                 continue
             if has_true_skipif(item):
-                print(f"skipping {item=} due to a true @skipif mark")
+                self._skip_because("skipif", item.nodeid)
```
Zac-HD, line 67: I think these markers have an optional `reason=` param/attribute; it'd be nice to expose that if it exists.

src/hypofuzz/dashboard.py
```diff
@@ -189,6 +210,7 @@ async def api_pycrunch_file(request: Request) -> Response:
     Route("/api/tests/", api_tests),
     Route("/api/tests/{node_id:path}", api_test),
     Route("/api/patches/", api_patches),
+    Route("/api/collected_tests/", api_collected_tests),
```
Zac-HD, line 213: (for later) this makes me think we should have some API documentation, or maybe just point users at a [schema](https://www.starlette.io/schemas/)?

Zac-HD (approved):
See `reason=`, then let's ship it!

tests/test_collection.py
tybug: oh thank god we can get rid of this lol

src/hypofuzz/interface.py
tybug: nice call. we're not actually displaying this extra info on the dashboard right now, but we will eventually

PR #69, opened 2025-03-22
Add ability to back a dashboard with a json file

tybug:
Adds a page at `/_dev` with a "download dashboard state" button. The downloaded json file can then be used with `hypothesis fuzz --dashboard-only --internal-backing-json-file` to start a dashboard using the data from that file. Intended for use at a future dashboard.hypofuzz.com.

Both the dev page and the flag are really only intended to be used by us, so I'm open to ideas of how to more clearly communicate that. Right now the `_dev` page is built for / accessible by everyone, it's just not linked anywhere on the sidebar, so a user will have no idea it exists unless they know to look for it. We could limit this page to being built on localhost, but tbh I don't see much harm in making a hidden page available to everyone. Who knows, maybe it'll help with debugging setups down the line.


Zac-HD:
I'm not convinced we need the `/_dev/` page, vs just GET-ing the `/api/backing_state/` json response?

More broadly, I was thinking that we could build a fully static frontend rather than having our Python backend serve from a json file (we could achieve the latter with no fuzzing workers just by copying the database, after all).  This is a bit more annoying due to the build etc, but not-operating-a-server seems worth it to me...

tybug:
Building a fully static frontend: this makes sense, I'm working on this. I'm thinking we have a `data.json` file which the dashboard prefers loading from, if present. The alternative is baking the data into the html of every page, which is more work to build.

tybug:
OK, dashboard now should work statically. I tested with a `--dashboard-only` server, but not from a static `index.html`, because chrome rejected some local requests for CORS.

> I'm not convinced we need the /_dev/ page, vs just GET-ing the /api/backing_state/ json response?

I mostly wanted one canonical and easy place to get this data when you or I updates the dashboard. I'd also be fine with a `scripts/get_backing_data.py` which makes this GET request to a currently-running dashboard

src/hypofuzz/entrypoint.py
```diff
@@ -71,13 +79,26 @@ def fuzz(
...
     This process will run forever unless stopped with e.g. ctrl-C.
     """
+    if internal_backing_json_file is not None and not dashboard_only:
+        raise click.UsageError(
+            "--internal-backing-json-file must be combined with --dashboard-only"
+        )
+
     dash_proc = None
+    if internal_backing_json_file is not None:
+        internal_backing_json_file = Path(internal_backing_json_file)
```
Zac-HD, line 89: mypy:
```
entrypoint.py:89: error: Incompatible types in assignment (expression has type "Path", variable has type "str | None")
```

Zac-HD (commented):
This is looking sweet!

Final bonus round, we currently have [this static dump](https://github.com/Zac-HD/hypofuzz/tree/master/docs/example-dashboard) which gets served from https://hypofuzz.com/example-dashboard/ - is it feasible to commit a `data.json`, and have the github-pages build set up the real static-data dashboard for us?

tybug:
Playing around with this in my fork. Unfortunately annoying to test because I have an extra /hypofuzz prefix `https://tybug.github.io/hypofuzz`. We might have to do some yoloing on master here...I'll pick it up tomorrow

Zac-HD:
If we can _avoid_ yolo-ing onto master, we could get the advantage of per-PR preview builds, as we have for Hypothesis' docs - and that would be pretty nice to have!  (though if it's a PITA, we yolo ofc)

tybug:
GHA can't write to both a domain and a subdomain in the same repo, so we can't quite match readthedocs, but I've tried a workflow with pushes both a docs and dashboard preview to `hypofuzz.com/preview-{pullid}`. It just pushes to a subdir on `gh-pages`, so we'll have to worry about cleanup (remove any `preview-*` dirs on master commit? But then this wipes the preview for other open PRs. Check mtime?)

.github/workflows/ci.yml
```diff
@@ -30,6 +30,17 @@ jobs:
         npm run format:check --prefix src/hypofuzz/frontend
 
   docs:
+    if: >
```
Zac-HD, line 33: looks like this is invalid syntax? https://github.com/Zac-HD/hypofuzz/actions/runs/14133098296/workflow#L33

tox.ini
```diff
@@ -19,7 +19,7 @@ deps =
     --requirement deps/docs.txt
 commands =
     pip install --editable .
-    sphinx-build -W --keep-going docs-src docs/docs {posargs}
+    sphinx-build -W --keep-going docs-src {env:HYPOFUZZ_DOCS_OUTPUT_DIR:docs/docs} {posargs}
```
Zac-HD, line 22: I think we'll also need to configure `passenv` for tox

.github/workflows/ci.yml
tybug: ack, I was wondering why the checks on this PR were waiting indefinitely, unfortunate ux that this is only accessible from the actions tab

tox.ini
tybug: great call thanks, this probably saved me a lot of frustrated debugging

tybug:
This approach might be dead at a permissions level. I don't think github supports adding the write permission to tokens for PRs from forks:

> You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the Send write tokens to workflows from pull requests option in the GitHub Actions settings. For more information, see [Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-private-repository-forks).

This quote makes it sound like it's possible for all repositories, but following the link gives:

> Send write tokens to workflows from pull requests - Allows pull requests from forks to use a GITHUB_TOKEN with write permission.

under the `Enabling workflows for forks of **private** repositories` header (emphasis mine) which implies this is only possible for private repos. Which seems possible as enabling write perms globally for a public fork is more dangerous than a private fork.

An alternative is that you and I start making branches from this repository, instead of forks...or maybe a manual-summon command like `/preview` could circumvent the fork permissions issue?

Zac-HD:
You're welcome to make PRs from branches-in-this-repo if you'd like; I already do.

I haven't looked into this enough to have strong opinions on the object-level tradeoffs, but my general goal here is to enable fast iteration with confidence and minimum manual labor each time, and so in-repo prs for all two active contributors seems like it might be a pretty good choice to me.

tybug:
going to close in favor of https://github.com/Zac-HD/hypofuzz/pull/70

PR #70, opened 2025-04-02
Dashboard previews and other misc improvements

tybug:
take two of https://github.com/Zac-HD/hypofuzz/pull/69, but branched off this repo to see if that resolves permissions issues


tybug:
fixes the permission issue! base url still isn't quite right, I will look into that

src/hypofuzz/entrypoint.py
```diff
@@ -54,6 +55,12 @@
     nargs=-1,
     metavar="[-- PYTEST_ARGS]",
 )
+@click.option(
+    "--internal-backing-json-file",
+    type=click.Path(exists=True),
+    help="Back the dashboard display using a static json file. Written for the "
+    "example dashboard on hypofuzz.com. This option implies --dashboard-only.",
+)
```
Zac-HD, line 63: I don't think we should expose this as a user-facing option _unless_ we think it's also a useful feature for users, which... _maybe?_  If you want to share the results to someone else who has Hypofuzz, but without the database, or you want to archive the state of a run at some historical point?  I can see an argument, actually, but we should then document the user-facing reasons to use this.

src/hypofuzz/entrypoint.py
```diff
@@ -71,13 +79,27 @@ def fuzz(
 
     This process will run forever unless stopped with e.g. ctrl-C.
     """
+    if internal_backing_json_file is not None and not dashboard_only:
+        raise click.UsageError(
+            "--internal-backing-json-file must be combined with --dashboard-only"
+        )
```
Zac-HD, line 85: docs say "implies", this is "errors without" - I think I'd be fine with either, but we should also error if you've got a backing json file and pass any other arguments (including unhandled args for forwarding to pytest)

tybug:
I'll probably be debugging live in this PR, so feel free to mute notifications - I'll ping when/if needed

tybug:
@Zac-HD alright this PR is ready for review: https://hypofuzz.com/preview-70/example-dashboard/

we should swap out the `dashboard_state.json` with a better one though, maybe leave hypothesis-jsonschema running overnight?

scripts/update_dashboard_state.py
```diff
@@ -0,0 +1,21 @@
+import argparse
+import json
+from pathlib import Path
+from urllib.parse import urlparse
+
+import requests
+
+parser = argparse.ArgumentParser()
+parser.add_argument("url")
```
Zac-HD, line 9: we could default this to `"http://localhost:9999/"`, since that's usually where it'll be

Zac-HD (approved):
Looks great!  Let's merge this to get our much nicer new dashboard onto the website, and then sure, I'd be happy to put up a more-carefully-chosen run as the example.

Zac-HD:
oops, the line in https://hypofuzz.com/preview-70/example-dashboard/#/tests/tests%2Ftest_provider.py%3A%3Atest_misaligned_kwargs goes off the top of the chart; I guess we're assuming that the maximum number of branches is from the latest report or something?

tybug:
yeah I expect that's what's happening as well, but the easy fix / investigation didn't work. On my shortlist!

PR #71, opened 2025-04-14
Use `sys.monitoring.events.BRANCH` for coverage collection on python 3.12+

tybug:
Also brings compatibility with new hypothesis versions.


tests/test_coverage.py
```diff
@@ -0,0 +1,309 @@
...
+
+    with Collector(f) as collector:
+        f(should_break=False)
+    branches = collector.branches
+
+    with Collector(f) as collector:
+        f(should_break=False)
+    # this assertion is descriptive, not perscriptive. We wish there was a new
+    # branch here.
+    assert collector.branches == branches
```
tybug, line 278: This was the only case I've found so far where `sys.monitoring` is not ideal.

Overall I've been pretty happy with how closely it matches what I expect. `sys.monitoring` deals in bytecode instructions, not in line/column numbers, so you have to adjust expectations sometimes; you might expect that this code:

```python
if a == 1:
    pass
b = 1
```

with `a = 0` has a branch from line 1 col 0 to line 3 col 0, but it really has a branch from line 1 col 4 (the `a` var) to line 3 col 4 (to LOAD 1 onto the stack, I think). The arcs are strictly internal, so I'm fine with this.

tybug:
A lot more failures than I was expecting, especially on 3.12, which passes cleanly for me locally. I'll look at what's going on here tomorrow

tests/test_coverage.py
Zac-HD: I note you've got `should_break=False` twice there.

Also iirc `coverage` deals with this somehow...

tybug: Thanks for catching that (it didn't change the conclusion, though). I'll look at how coverage does this, but I wouldn't be too surprised to find the answer is a manual bytecode analysis

tybug:
failures on 3.12+ were from interpreter bytecode differences in minor versions. I reproduced locally after updating from 3.12.3 to 3.12.10, and updated the tests

tests/test_coverage.py
tybug: It might be fine that this doesn't count as a branch actually? If `break` executes in one iteration and not another, it must have been because some other branch was taken in one and not the other, and that branch would distinguish this break behavior. 

Exceptions/returns might be nonobservable control flow here though

Zac-HD: Could get those via CALL / RETURN / {PY_,C_}UNWIND events maybe?  Seems fine to revisit that later though.

Zac-HD (approved):

Issue #67, opened 2025-03-20
Improve table UX


tybug:
* Should we ditch the idea of "table segments" (failing / started executing / etc) and move that to a column, maybe with a colored pill to make it stand out? That would make  sorting easier.
* Every column should be sortable (only within its segmented column? see above)
* Add a search/filter box for rows
  * I expect this to be used most often for node ids, so it could be nice to accept basic regex as well

PR #73, opened 2025-04-15
Improve table UI

tybug:
Part of #67. I've dropped the notion of "table segments" in preference of a status pill column, and in preparation for filtering/sorting (coming soon).

Preview:

Image

Image

### DIFF
--- a/src/hypofuzz/database.py
+++ b/src/hypofuzz/database.py
@@ -1,4 +1,5 @@
 import json
+from base64 import b64decode, b64encode
 from collections.abc import Iterable
 from functools import cache
 from typing import Any, Optional, TypedDict
@@ -36,6 +37,7 @@ class Report(TypedDict):
 
 
 class Metadata(TypedDict):
+    database_key: str
     nodeid: str
     seed_pool: list[list[str]]
     failures: list[list[str]]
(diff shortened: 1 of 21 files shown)


src/hypofuzz/frontend/src/utils/prototypes.tsx
```diff
@@ -0,0 +1,27 @@
...
+      const keyB = key(b)
+
+      for (let i = 0; i < Math.min(keyA.length, keyB.length); i++) {
+        if (keyA[i] < keyB[i]) return -1
+        if (keyA[i] > keyB[i]) return 1
+      }
+      return 0
+    })
+  }
+}
```
tybug, line 27: Adding a new function to the prototype might be too magic, but I find this so much more ergonomic than the two-element `.sort` in javascript

src/hypofuzz/database.py
```diff
@@ -36,6 +37,7 @@ class Report(TypedDict):
 
 
 class Metadata(TypedDict):
+    database_key: str
     nodeid: str
     seed_pool: list[list[str]]
     failures: list[list[str]]
```
tybug, line 43: This is a breaking change for existing databases. We should work out how to do migrations at some point, but I think we have a while yet before we stabilize the db to a point where that becomes worthwhile

(I'm using `database_key` as the unique key for react's rendering, and in general we've discussed before that `database_key` should be the backing unique key while node id is for display only)

Zac-HD: See https://github.com/Zac-HD/hypofuzz/issues/3#issuecomment-2808734642

- We can, in general, safely drop basically anything from the database except for our saved inputs.  It costs us pretty graphs and some state estimation which makes the fuzzer less efficient for... minutes to hours (?), but it's recoverable.  
  - We should test that we're OK if the database has unexpected data shapes in it though; and "parse don't validate" suggests loading as frozen dataclasses rather than dicts - that's a natural place to put migration logic to make new objects out of older saved data.
- there are enough database changes in that comment that I suggest we do them first, then rebase these frontend PRs.  Sorry!  Note that we can just show the latest run for now, and do aggregation later.

PR #74, opened 2025-04-15
Add ability to sort tables 

tybug:
- [x] depends on #73 

Preview:

image


Zac-HD (approved):
approving to unblock.

(I went to check the preview dashboard view, but that only works from within-repo branches - want to start using those?  we can always prefix `zac/branchname` / `liam/branchname`)

PR #75, opened 2025-04-16
Add proper database classes and store more data

tybug:
Part of https://github.com/Zac-HD/hypofuzz/issues/3#issuecomment-2808734642

### DIFF
--- a/src/hypofuzz/hypofuzz.py
+++ b/src/hypofuzz/hypofuzz.py
@@ -3,16 +3,20 @@
 import contextlib
 import itertools
 import os
+import platform
 import socket
 import sys
 import time
 import traceback
+from base64 import b64encode
 from collections.abc import Callable, Generator
 from contextlib import suppress
 from functools import lru_cache
 from random import Random
-from typing import Any, Optional, Union, cast
+from typing import Any, Optional, Union
+from uuid import uuid4
 
+import hypothesis
 from hypothesis import settings
 from hypothesis.control import BuildContext
 from hypothesis.core import (
@@ -38,6 +42,7 @@ from hypothesis.reporting import with_reporter
 from hypothesis.vendor.pretty import RepresentationPrinter
 from sortedcontainers import SortedKeyList
 
+import hypofuzz
 from hypofuzz.corpus import (
     BlackBoxMutator,
     CrossOverMutator,
@@ -46,7 +51,7 @@ from hypofuzz.corpus import (
     get_shrinker,
 )
 from hypofuzz.coverage import CoverageCollectionContext
-from hypofuzz.database import Metadata, Report, WorkerT, get_db
+from hypofuzz.database import Metadata, Phase, Report, WorkerIdentity, get_db
 from hypofuzz.provider import HypofuzzProvider
 
 record_pytrace: Optional[Callable[..., Any]]
@@ -54,6 +59,8 @@ try:
     from hypofuzz.pytrace import record_pytrace
 except ImportError:
     record_pytrace = None
+
+process_uuid = uuid4().hex
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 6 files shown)


src/hypofuzz/hypofuzz.py
```diff
@@ -384,49 +391,56 @@ def _replace_metadata(self, metadata: Metadata) -> None:
...
+    def _phase(self) -> Phase:
+        phase = Phase.GENERATE
+        if self._replay_queue:
+            phase = Phase.REPLAY
+        elif self.pool._in_distill_phase:
+            phase = Phase.DISTILL
+        elif self.pool.interesting_examples:
+            phase = Phase.SHRINK if self.shrinking else Phase.FAILED
+        return phase
+
```
tybug, line 404: I went for `phase` over `how_generated` because a report is an aggregate unit, so I want to make claims about the current phase, not that all examples in the report were generated a particular way

tybug: oh, but they *are* all generated the same way, because we emit a report whenever we switch phases. This is what lets us linearize. I will probably go back to `how_generated`

tybug:
* Linearized-history view is currently the only dashboard view, option to select runs-in-last-n-days coming later
* Big PR; important bits are `def linearize_reports` and the dataclasses in `database.py`

tests/test_linearize.py
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+
+@st.composite
+def reports(
+    draw, *, count_workers: Optional[int] = None, overlap: bool = False
+) -> list[Report]:
+    # all of this min_size=len(uuids) etc is going to lead to terrible shrinking.
+    # But the alternative of while draw(st.booleans()) will generate too-small
+    # collections. Use `more` from hypothesis internals?
```
Zac-HD, line 42: Let's decompose this: generate a `worker_and_run()` or something, and then have `reports()` draw a list of those.  We can pass in a list of (db_key, nodeid) pairs for each report to sample from; and have the multi-worker reports strategy add whatever offset we need after the fact if overlap=False (noting that we probably want a smaller upper bound on timestamps).

src/hypofuzz/hypofuzz.py
```diff
@@ -465,8 +472,20 @@ def fuzz_several(*targets_: FuzzProcess, random_seed: Optional[int] = None) -> N
...
-    """Return a json blob identifying the machine running this code.
+def _git_head() -> Optional[str]:
+    try:
+        return subprocess.check_output(
+            ["git", "rev-parse", "HEAD"],
+            timeout=10,
+            text=True,
+        ).strip()
+    except Exception:
+        return None
```
Zac-HD, line 483: I'd like to handle the case where you run `hypothesis fuzz -- path/to/repo/tests/`.  Probably the way to do that is to make this a function of `Path -> str | None`, and have each worker run it for the directory containing the test file?  This also complicates `WorkerIdentity` a bit, since a given process could be fuzzing tests from multiple git repos.

Happy to defer that to a follow-up PR though.

Zac-HD (commented):
looking good!

src/hypofuzz/hypofuzz.py
tybug: I'm not too worried about the multiple-repos case; worst case we include the repository name or path and ignore our git heuristics if not in the same repo.

tests/test_linearize.py
tybug: ok if I come back to this one? I definitely want a strong strategy here, but also want to write the overlapping case first before making it smarter

Zac-HD (approved):
lgtm

Zac-HD:
My bad for the hasty review, seriously, but this implementation is wrong once you have a restart - sorting by elapsed time will interleave all the runs, and then crash with assertion errors.

Having played around a bit in a branch (https://github.com/Zac-HD/hypofuzz/compare/zac/linearize), I think the solution is to change `REPORTS` to be a nested dictionary: `nodeid -> worker_uuid -> list[Report] sorted by timestamp`

- With that structure, it's easy to derive the diffs just be iterative over the elements of each list
  - then linearize by concatenating all the per-worker-uuid lists and sorting by typestamp, and dropping not-at-the-start replay entries.
  - I actually think that computing diffs in the dashboard server kinda sucks; either we should take the space hit and put that in the database (ie denormalize a bit; our worker-identity mapping is already heavy-ish), or do it in the frontend.
- we need all those individual lists anyway, since we want an option to plot them separately on the per-test pages
- also, in every place we construct a `Report` or a `Metadata` loaded from the database, we need to handle 'parsing errors' due to invalid json, or missing/extra keys due to writes from an older or newer fuzzer version.
  - gosh that's going to be annoying, ugh, at least it's not too many places...
  - I think we just skip over those entries; I'm cautious about deleting stuff (imagine you update the fuzzer, have to roll back, and in the meantime we dropped all your metadata...) - we can do something sensible later.

PR #76, opened 2025-04-22
Remove references to pytrace

tybug:


Zac-HD (approved):
Ah, even after we discussed this, seeing it go makes me a little melancholy.  Time-travel debugging - with antithesis-style multiverse, and Scrutineer annotations - is still part of the dream... but this implementation wasn't there yet.  We'll be back someday, when the time is right and the tech is ready.

🫡

PR #77, opened 2025-04-23
Fix example dashboard build

tybug:
no guarantees, but let's try this

apparently in vite `--base=//example-dashboard` != `--base=/example-dashboard`


tybug:
wrong repo ack

PR #79, opened 2025-04-23
Add ability to filter table rows

### LINKED ISSUE #67: Improve table UX

tybug:
Closes #67

I'm not particularly happy with the placement of the search box, but I don't have any better ideas!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,19 +76,17 @@ jobs:
         cp docs-src/dashboard_state.json docs/${BASE}/example-dashboard/assets/dashboard_state.json
     - name: Commit docs
       run: |
-        git config --local user.email "action@github.com"
-        git config --local user.name "GitHub Action"
-
         cp -r docs/${BASE}/docs $RUNNER_TEMP
         cp -r docs/${BASE}/example-dashboard $RUNNER_TEMP
 
-        git add --force docs/${BASE}/docs
-        git add --force docs/${BASE}/example-dashboard
         git fetch --all
         git switch -c gh-pages origin/gh-pages
 
         cp -r $RUNNER_TEMP/docs docs/${BASE}
         cp -r $RUNNER_TEMP/example-dashboard docs/${BASE}
+
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
         git commit -m "build docs (base=${BASE})"
         git push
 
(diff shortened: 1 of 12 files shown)


tybug:
Preview is up: https://hypofuzz.com/preview-79/example-dashboard/ (getting a side-by-side comparison with https://hypofuzz.com/example-dashboard/ is so nice for review)

Github actions code looks okay - we didn't overwrite the actual /example-dashboard/ this time - though the real test will be making sure the example dashboard updates on master when merged.

tybug:
Iterative preview builds in action: https://github.com/Zac-HD/hypofuzz/commit/13fdf9e2e4128af13b8a2ec239b4936951892b61, looks all good, though running on master is still untested

Zac-HD:
quick notes:
- meta note, wow yeah having an online preview is great, not just for convenience but to be confident that I'm seeing the same thing as you.  10/10 useful feature!
- can we have the search box in the same vertical space as the `Tests` heading, at least when the screen is wide enough?
- search-box test suggestion "Filter by test name", to make it clear that it's only on that attribute (and status?)
- (since you mentioned mobile, maybe for later) I noticed that the test-name column gets very very narrow as the table width decreases.  Maybe we should switch to a card-like layout for narrow windows, where the test name sits over the other attributes in each row, but we maintain the same table header?
- when we've filtered, I'd add a note to the bottom of the table, "showing NN of MMM tests `[clear filter x]`", and an `x` to clear the search box
- when I drag the window narrower, there's almost always some stuff off to the right of the screen, even as the sidebar comes and goes.

tybug:
* I don't love the look of it being flush with the header, but I've put it about halfway up here, because I agree it takes up too much vertical space otherwise. You can set `margin-top: -30px` on the filter box to see what it would look like flush.
* it's by status as well though, as you noted. (though that didn't work until this push, oops). I think I prefer just `Filter`, since all textual columns should be searchable
* oh this could be a nice layout for mobile, I'll try it (later)
* done
* yeah, best-effort for now. Can improve in the future

Zac-HD (approved):
Looks great!

PR #82, opened 2025-04-25
Table row alignment and graph height bugfix

tybug:


Zac-HD (approved):

tybug:
I was trying to add a fancy clickable url in the actions status like the readthedocs action has, but seems it won't be trivial. I'll try that another time

PR #83, opened 2025-04-25
Include parametrization in db key

tybug:
We weren't doing this before, either because the manual invocation of pytest doesn't run other plugins, or because the hypothesis plugin adds the key at `pytest_runtest_call` and not at collection time.

### DIFF
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -4,7 +4,7 @@ HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 
 ## 25.04.9
 
-Add a search bar to the coverage graph, and fix failures for parametrized tests being written to the incorrect database key.
+Also filter the coverage graph when filtering the test table, and fix failures for parametrized tests being written to the incorrect database key.
 
 ## 25.04.8
 
--- a/docs-src/changelog.md
+++ b/docs-src/changelog.md
@@ -2,6 +2,10 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 25.04.9
+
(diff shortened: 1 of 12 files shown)


tybug:
direct link to the preview, pretty cool: image

tests/test_database.py
```diff
@@ -33,3 +33,24 @@ def test_database_stores_reports_and_metadata_correctly():
...
+    db = HypofuzzDatabase(DirectoryBasedExampleDatabase(db_dir))
+    assert not list(db.fetch(b"hypofuzz-test-keys"))
+
+    with fuzz(test_path=test_dir, n=2):
+        # this will time out if the test keys are the same across parametrizations
+        wait_for(
+            lambda: len(list(db.fetch(b"hypofuzz-test-keys"))) == 2,
+            timeout=10,
+            interval=0.1,
+        )
```
Zac-HD, line 56: I'd like to _also_ test that the test keys used by the fuzzer are the same two keys that would be used by running pytest+hypothesis over the same test code.

docs-src/changelog.md
```diff
@@ -2,6 +2,10 @@
 HypoFuzz uses [calendar-based versioning](https://calver.org/), with a
 `YY-MM-patch` format.
 
+## 25.04.9
+
+Improve uniqueness of database keys for tests which use `@pytest.mark.parametrize`.
```
Zac-HD, line 7: It's not just uniqueness; it's ensuring that Hypofuzz uses the same db keys as Hypothesis, so that replay works!

tybug:
I've added a search bar to the coverage graph as well (just noticed: need to hide for the single-test case)

Zac-HD (approved):
Looks good, but can we run the filter for both table and graph off a single input box?  I'd like to keep them in sync.  (then merge 😁)

PR #90, opened 2025-04-29
Various improvements

tybug:
Using https://prismjs.com/ for syntax highlighting, currently only for failures:

image

I went with prism over https://highlightjs.org/ since prism claims to be a much smaller bundle size, which seems to check out, though I didn't actually test highlight js.


Zac-HD:
makes sense; I mentioned highlightjs because it's got much better `.patch` support - worth the larger bundle IMO.

tybug:
yeah that is way better, nice. Also only negligibly larger (~2kb) once I told it to only import the langages we're using.

image

Zac-HD (approved):

Issue #63, opened 2025-03-08
Integrate Tyche with the dashboard


tybug:
[Tyche](https://github.com/tyche-pbt/tyche-extension) is a tool for visualizing property-based tests, supported by Hypothesis via the [observability api](https://hypothesis.readthedocs.io/en/latest/observability.html). Since Tyche is "just" a react app, it would be pretty cool to embed Tyche results for each test in the Hypofuzz dashboard. 

This may require tweaking metrics for the browser, rather than the editor, as well as considering the history over which we display the observability results and where we store them (inside the dabatase `Metadata`?) Per-test-case insight is probably too much data to store, but we could aggregate in some way.

Zac-HD:
We clearly can't collect observability input for all inputs (...by default?) for performance reasons, so the real question is how to downsample.  A few possible answers:

- observe the covering corpus.  Ensures we see interesting and diverse behaviors, but is far from representative - for example, there are no invalid inputs in this set, even if they're common overall!  Requires re-executing these inputs but we want to do that anyway (#5).
- observe random inputs, at an exponentially-decaying frequency so that we approach a reasonable constant corpus size in the limit.  Unfortunately, this means that later parts of the run are wildly under-represented; while statistically "nothing happens there" it's also the most important part for users to understand in terms of e.g. performance optimizations.
- observe random inputs, at a low but roughly constant rate, and randomly evict older observations to maintain a constant-size set.  This gives a sample with density proportional to recency, which seems like a pretty nice property to maintain?
- All of the above, and let users toggle between them?

also cc @hgoldstein95, because I suspect you'll have good opinions on this topic ✨

hgoldstein95:
I think the key question is figuring out why a developer would use Tyche with hypofuzz. I think the most obvious story to tell is the one where we observe the covering corpus. In that situation, the user is trying to understand the input-space coverage for the best inputs that the fuzzer found — if the user finds that the covering corpus is missing a key part of the input space, that's actionable information that they can use to conclude that something was missing from their testing.

One could imagine a different story where the user is trying to debug the fuzzing setup itself, improve mutation, etc. and in those cases they may indeed want to know things about the invalid inputs. But in that case I think we may actually just want to let the user log all inputs. I think any attempt to down-sample will just be confusing: the fuzzer isn't drawing from a static distribution, so it's basically impossible to accurately understand the full set of inputs from some sample.

[...]

PR #91, opened 2025-04-30
Add observability reports

tybug:
- [x] depends on https://github.com/HypothesisWorks/hypothesis/pull/4379

Changes from https://github.com/Zac-HD/hypofuzz/tree/zac/db-metadata:

* changed `got: list` to a new `observed: Value` class; I assume the list was being used as a one-element pass-by-reference
* removed `self._observations_queue` in `FuzzProcess` (never accessed, handled by db instead)
* (as discussed) use `StateForGivenExecution`. This worked quite well, and let us get rid of our copied `RepresentationPrinter` code.

Observability can take up a substantial amount of space, it's basically the same as adding 300 covering examples to every test. The most substantial part is the `representation` string of the test function and the arguments. My `.hypothesis` stabilized at 48mb when running on hypofuzz. 

(actually - it's worse than adding 300 covering examples, because covering examples are distilled and so will have smaller representations, while observability examples are randomly sampled)


Zac-HD:
- changes from my branch all look good
- hmm, ~five mb per test _is_ a fair bit of data.  I'm inclined to leave it as-is for now, since I think that having good observability is _really really valuable_ - being able to dig in to raw-ish data reliably turns up important insights, and you don't need to do that very often to pay for the storage many times over.
  - but back-of-mind, we can always reduce the buffer size later - 300 is just a nice number for decent statistics, and five minutes a decent duration.  (I may also have, uh, skewed opinions on what counts as expensive here)
  - plausible option, we could consider compressing observations with gzip or zstd before storing them in the database.  Conveniently, this is easy to handle compatibly by checking the leading bytes for a "magic number" specifying the format, and treating `{"` as the magic number for uncompressed json; "try several and keep shortest" is a neat compression trick for this application.
- seems like we might want to update the backing json blob for the online example once we have observations?  I'm assuming that we'll want to get them onto the frontend in this PR too I suppose.
  - note that I'm not talking about #63, "just" ensuring that we no longer store the Metadata json in the database (instead loading both covering and failing examples from the stored observations)
- I had to manually `npm install highlight.js` to get the build working locally, maybe a followup for our backend?

I'll review again when you tag me.

src/hypofuzz/dashboard.py
```diff
@@ -292,20 +364,49 @@ def send_nowait_from_anywhere(msg: ListenerEventT) -> None:
...
-            # elapsed_time, when we have Metadata extend Report
-            METADATA[node_id] = metadata[0]
+    for fuzz_target in COLLECTION_RESULT.fuzz_targets:
+        # a fuzz target (= node id) may have many database keys over time as the
+        # source code of the test changes. Show only reports from the latest
+        # database key = source code version.
+        #
+        # We may eventually want to track the database_key history of a node id
+        # and at least transfer its covering corpus across when we detect a migration,
+        # as well as possibly showing a history ui in the dashboard.
```
Zac-HD, line 374: Hmm, maybe `hypofuzz-test-keys` is the mechanism which would let us do this?  i.e. collect by nodeid as here, then load reports from any other test-keys, and group them in if the nodeid was from a collected test.

Happy to leave it for a follow-up though.

src/hypofuzz/database.py
```diff
@@ -41,9 +51,68 @@ class WorkerIdentity:
...
+    call_repr: str
+    reproduction_decorator: str
+
+    @staticmethod
+    def from_json(data: dict) -> Optional["FailureRepresentation"]:
+        data = dict(data)
+        try:
+            return FailureRepresentation(**data)
+        except Exception:
+            return None
```
Zac-HD, line 114: as discussed: we can have Hypothesis store the `@reproduce_failure(...)` string in the observations, and store the traceback for _all_ failures, at which point we won't need this separate class.

src/hypofuzz/hypofuzz.py
```diff
@@ -37,61 +36,70 @@
...
-    # to imagine an intentionally-deeply-recursive use of this code.
-    assert depth <= 1000, (
-        f"Hypothesis would usually add {recursion_limit} to the stack depth of "
-        f"{depth} here, but we are already much deeper than expected.  Aborting "
-        "now, to avoid extending the stack limit in an infinite loop..."
+def reproduction_decorator(choices: ChoicesT) -> str:
+    """Return `@reproduce_failure` decorator for the given choices."""
+    return (
+        f"@reproduce_failure({hypothesis.__version__!r}, {encode_failure(choices)!r})"
     )
```
Zac-HD, line 73: as above, I'd like to handle this in Hypothesis

tests/test_corpus.py
```diff
@@ -114,5 +111,31 @@ def test_interesting_results_are_added_to_database(args):
...
+    for count in range(1, 10):
+        process._run_test_on(
+            ConjectureData.for_choices([1]), collector=Collector(test_a)
+        )
+        # we keep incrementing arc counts if we don't find new coverage
+        assert process.corpus.branch_counts == {((3, 11), (3, 11)): count}
+
+    process._run_test_on(ConjectureData.for_choices([2]), collector=Collector(test_a))
+    # our arc counts should get reset whenever we discover a new branch
+    assert process.corpus.branch_counts == {
```
Zac-HD, line 134: FYI: we do reset branch counts (soon behavior counts) every time we find a new branch (...behavior), and I think that's the right call at the moment.  Eventually we'll want to use a fancier estimator which can leverage information from the slightly-different prior distribution; iirc Boehme et al studied that at some point.

tests/test_fuzz_process.py
```diff
@@ -25,21 +27,23 @@ class CustomError(Exception):
...
     assert fp.status_counts[Status.INTERESTING] >= 1
-    (call_repr, _, _, tb_repr), *rest = fp._metadata.failures
-    assert not rest  # expected only one failure
-    assert tb_repr.endswith("test_fuzz_process.CustomError: x=1\n")
-    assert call_repr == "failing_pbt(\n    x=1,\n    y=0,\n)"
+    failures = list(db.fetch_failures(fp.database_key))
+    assert len(failures) == 1
+    representation = db.fetch_failure_representation(fp.database_key, failures[0])
+    assert "CustomError" in representation.traceback
+    assert representation.call_repr == "test_fails(\n    x=1,\n    y=0,\n)"
```
Zac-HD, line 49: This explicitly _doesn't_ have the explain-phase output that we're notionally testing for!

...which I realize is the status quo of the test, oops, but we'll want to fix that at some point 😅

src/hypofuzz/corpus.py
```diff
@@ -81,44 +87,47 @@ def sort_key(nodes: Union[NodesT, ConjectureResult]) -> tuple[int, tuple[int, ..
...
-        self._key = key
+        self._database_key = database_key
+        self._db = HypofuzzDatabase(hypothesis_database)
+        self._hypothesis_db = hypothesis_database
 
-        # Our sorted pool of covering examples, ready to be sampled from.
-        # TODO: One suggestion to reduce effective pool size/redundancy is to skip
+        # Our sorted corpus of covering examples, ready to be sampled from.
+        # TODO: One suggestion to reduce effective corpus size/redundancy is to skip
         #       over earlier inputs whose coverage is a subset of later inputs.
```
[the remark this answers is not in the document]
tybug: I wonder if we want to present two types of patches, "covering behaviors" and "covering branches"? Seems like "covering behaviors" might be substantially larger, and a developer might want to pare down the applied patch to just branch coverage

tests/test_corpus.py
tybug: sounds good! This test isn't really related to this PR, I was just trying to understand the corpus logic (because I thought it had a bug, and then discovered it didn't).

src/hypofuzz/corpus.py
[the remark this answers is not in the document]
Zac-HD: Sounds good, but to do that we need to track which inputs cover novel branches vs only behaviors, and that's pretty annoying to deal with.  We might even want to deal with the set-cover problem to show fewer inputs again... I suggest a tracking issue for later.

tests/test_fuzz_process.py
```diff
@@ -25,21 +27,23 @@ class CustomError(Exception):
...
     assert fp.status_counts[Status.INTERESTING] >= 1
-    (call_repr, _, _, tb_repr), *rest = fp._metadata.failures
-    assert not rest  # expected only one failure
-    assert tb_repr.endswith("test_fuzz_process.CustomError: x=1\n")
-    assert call_repr == "failing_pbt(\n    x=1,\n    y=0,\n)"
+    failures = list(db.fetch_failures(fp.database_key, shrunk=True))
+    assert len(failures) == 1
+    observation = db.fetch_failure_observation(fp.database_key, failures[0])
+    assert "CustomError" in observation.metadata["traceback"]
+    assert observation.representation == "test_fails(\n    x=1,\n    y=0,\n)"
```
Zac-HD, line 49: Per the diff this is a pre-existing problem, but this test shows that we're _not_ getting annotations from the explain phase, and we should follow up to fix that - fuzzing is the perfect time to spend more compute for a more useful message!

Zac-HD (approved):
🎊 

I'll give you a change to review my late-night commits; then merge when you're happy.  We'll also need to make sure we collect all the followup comments into an issue later 😅

PR #96, opened 2025-05-08
Split observations for faster load time

tybug:
* for local dashboard websocket: split initial message into "tests" and "observations", loaded asyncly
* for example dashboard: separate `backing_state` files. Overview page reads `tests.json` and each individual test page reads both `tests.json` and `observations.json`, async-ly (using a web worker, since json.parse is sync).
  * we may want to split this into per-test observation files for faster load times
* Only look for dashboard_state if requested at `npm run build`-time with an envvar


Zac-HD (approved):
Looks good!

When viewing on a slow device though, I do notice that the covering examples are missing for ~10s while everything loads - so I think we should have a very simple *`{thing} loading...`* message from page load until we actually pop in the card.

PR #97, opened 2025-05-10
Refactor report linearization

tybug:
The concrete decision I'm building outwards from is that both the frontend and the backend need to be able to linearize reports, since we want to filter by worker on the frontend. This linearization now happens initially on `Test` init, and incrementally in `add_report`,  in both python and ts.

We can either send the frontend `reports: list[Report]` or `reports_by_worker: dict[str, list[Report]]`. I've done the latter since the frontend would just transform the former into the latter anyway. (note that in either case we need to send the unlinearized reports to the dashboard, else it could not re-linearize when filtering). 

Grouping by worker also lets us drop `report.worker` from each report and put it in `reports_by_worker: dict[str, tuple[WorkerIdentity, list[Report]]]`, for substantial network/parsing savings. That's not yet done in this PR though.


src/hypofuzz/dashboard.py
```diff
@@ -54,84 +54,154 @@
...
-    reports: list[Report]
-    reports_offsets: ReportOffsets
     rolling_observations: list[Observation]
     corpus_observations: list[Observation]
     failure: Optional[Observation]
+    reports_by_worker: dict[str, list[Report]]
+
     status_counts: StatusCounts = field(init=False)
     elapsed_time: float = field(init=False)
+    linear_reports: list[Report] = field(init=False)
```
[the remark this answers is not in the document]
tybug: I've done this almost verbatim, except:

- unordered reports are dropped rather than recomputed (I expect this to be rare, have left a TODO)
- use optional attributes in `Report` instead of a `ReportWithDiff` subclass. I think this will generalize better and avoid a mixin-based mess 
- no LRU cache js-side yet
- phase=replay reports are simply dropped from `linear_reports`. We could cache +0 for these, but then they would still appear on the graph. I considered a `include_replay: bool = False` parameter, but haven't added it since we don't actually need `include_replay=True` anywhere yet.

src/hypofuzz/dashboard.py
```diff
@@ -49,89 +50,242 @@
...
+                    running += value
+                    cumsum.append(running)
+                cache[since] = (start_idx, cumsum)
+
+            return cumsum
+
+        cumsum = []
+        start_idx = bisect.bisect_right(
+            [report.timestamp_monotonic for report in self.linear_reports], since  # type: ignore
+        )
```
Zac-HD, line 245: Constructing a temporary list like this is O(n), can we use the `.bisect_right()` method on the `self.linear_reports` SortedKeyList instead?

Zac-HD (commented):
quick review, back later

src/hypofuzz/database.py
```diff
@@ -150,15 +153,26 @@ class Report:
...
+    # This fields are not stored in the database, but are computed for an
+    # individual Report, relative to another Report.
+    status_counts_diff: Optional[StatusCounts] = field(default=None)
+    elapsed_time_diff: Optional[float] = field(default=None)
+    # The timestamp of consecutive reports is not always monotonic - due to
+    # daylight savings time, for example, or merely NTP clock updates. We therefore
+    # store a separate timestamp_monotonic, which uses timestamp if the timestamp
+    # is monotonic relative to the previous report, and otherwise
+    # previous_report.timestamp_monotonic + elapsed_time_diff.
+    timestamp_monotonic: Optional[float] = field(default=None)
```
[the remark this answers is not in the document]
tybug: Slightly disagree, but I think `ReportWithDiff` is fine until/if we add more `Report` subclasses

tests/test_linearize.py
```diff
@@ -85,8 +85,9 @@ def reports(
...
                 st.builds(
                     Report,
-                    database_key=st.text(st.characters(codec="ascii")),
-                    nodeid=st.text(st.characters(codec="ascii")),
+                    # TODO draw consistent database_key and nodeid per-worker
+                    database_key=st.just(b""),
+                    nodeid=st.just(""),
                     elapsed_time=st.just(elapsed_time),
                     timestamp=st.just(timestamp),
                     worker=workers(uuid=uuid),
```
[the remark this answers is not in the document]
tybug: It's tricky because reports in a worker guarantee that `elapsed_time` and `status_counts` both increase monotonically. If just one attribute increased, I would generate `st.lists(Report).map(partial(sorted, key=lambda r: r.timestamp))`. But with two attributes, I can't generate `st.lists(Report)` ahead of time, since the ordering is partial.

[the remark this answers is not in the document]
Zac-HD: oof, yeah.

The usual idiom is to cache what you've generated before, in this case via a `st.shared(st.builds(dict[uuid, SortedKeyList[Report]]))`.  But gosh that bisection is going to be annoying if we don't want to generate invalid status_counts...

no, it's actually not that bad: we can generate whatever we want, bisect to find the position by elapsed_time, and then clamp the status_counts to be legal (or reject if if we've dropped in between a previous and next report that differ by exactly one test case)

still unclear that it's worth doing in this specific PR, just open a follow-up issue if not.

src/hypofuzz/database.py
```diff
@@ -174,6 +177,20 @@ def from_json(encoded: bytes, /) -> Optional["Report"]:
...
+    # This fields are not stored in the database, but are computed for an
+    # individual Report, relative to another Report.
+    status_counts_diff: StatusCounts
+    elapsed_time_diff: float
+    # The timestamp of consecutive reports is not always monotonic - due to
+    # daylight savings time, for example, or merely NTP clock updates. We therefore
+    # store a separate timestamp_monotonic, which uses timestamp if the timestamp
+    # is monotonic relative to the previous report, and otherwise
+    # previous_report.timestamp_monotonic + elapsed_time_diff.
+    timestamp_monotonic: float
```
Zac-HD, line 191: - I misspoke; daylight savings doesn't affect unix time, though NTP updates (or manual clock adjustments) do.
- it's not monotonic _relative to previous report_, but rather prev_report + elapsed_time diff.

```suggestion
    # Clock updates due to e.g. NTP can make time.time() non-monotonic, so
    # to preserve ordering in edge cases we define `timestamp_monotonic` as
    # `min(time.time(), previous_timestamp_monotonic + elapsed_time_diff)`.
    timestamp_monotonic: float
```

src/hypofuzz/hypofuzz.py
```diff
@@ -366,21 +376,35 @@ def _run_test_on(
...
+        # status_counts and elapsed_time have to be added to at the same time, without
+        # saving a report between them, or we might violate monotonicity invariants.
+        self.elapsed_time += time.perf_counter() - start
         assert observation.value is not None
         if self.corpus.add(data.as_result(), observation=observation.value):
-            self.since_new_cov = 0
+            # TODO this is wrong for Status.INTERESTING examples (that are smaller
+            # than the previous example for this interesting origin), which are
+            # successfully added to the corpus but don't represent a new branch.
+            self.since_new_branch = 0
```
Zac-HD, line 387: 🗒️ noting for later followup

src/hypofuzz/hypofuzz.py
```diff
@@ -477,7 +502,7 @@ def has_found_failure(self) -> bool:
 def fuzz_several(*targets_: FuzzProcess, random_seed: Optional[int] = None) -> None:
     """Take N fuzz targets and run them all."""
     random = Random(random_seed)
-    targets = SortedKeyList(targets_, lambda t: t.since_new_cov)
+    targets = SortedKeyList(targets_, lambda t: t.since_new_branch)
 
     # Loop forever: at each timestep, we choose a target using an epsilon-greedy
     # strategy for simplicity (TODO: improve this later) and run it once.
```
Zac-HD, line 508: 🗒️ noting for later: we should flush a report when we switch targets (but avoid writing exact-duplicate reports)

tests/test_linearize.py
```diff
@@ -61,6 +65,10 @@ def reports(
     for uuid, (start_time, end_time) in zip(uuids, intervals):
         # reports from the same worker always have monotonically increasing
         # coverage, timestamps, and elapsed_time
+        # (TODO: we don't actually rely on timestamps being ordered. But our logic
+        # for unordered reports isn't correct yet either (we drop them, when we
+        # should correctly recompute/reinsert). Drop the .map(sorted) for timestamps
+        # when we implement this correctly.
```
Zac-HD, line 71: 🗒️ noting for later followup

tox.ini
```diff
@@ -30,7 +30,7 @@ deps =
     pytest7: --requirement deps/test-old.txt
 commands =
     pip install --editable .
-    pytest {posargs:-n auto}
+    pytest {posargs:-n auto} -vvv
```
Zac-HD, line 33: stray debug code?  You should be able to call `tox -e test -- -vvv` iirc.

src/hypofuzz/dashboard.py
```diff
@@ -50,89 +52,249 @@
...
 
     def add_report(self, report: Report) -> None:
-        status_counts = self.reports_offsets.status_counts
-        elapsed_time = self.reports_offsets.elapsed_time
-        counts_diff = report.status_counts - status_counts.setdefault(
-            report.worker.uuid, StatusCounts(dict.fromkeys(Status, 0))
+        # we use last_report to compute timestamp_monotonic, and last_report_worker
+        # to compute status_count_diff and elapsed_time_diff
+        last_report = self.linear_reports[-1] if self.linear_reports else None
+        last_report_worker = (
```
Zac-HD, line 143: Conceptually, we should use `last_report_worker` for all attributes including `timestamp_monotonic`.  That does require the ability to insert not just append, so we can defer it for now.

Zac-HD (approved):
Looks great; merge after any minor comments you want to fix, and then it'll be easier to review a follow-up PR later.

tox.ini
tybug: Trying to track down an infinite-hanging test on CI, but I think I've narrowed down *which* test is causing this at least.

src/hypofuzz/database.py
tybug: (max, right, not min?)

src/hypofuzz/dashboard.py
Zac-HD: It only breaks monotonicity if you append to `linear_reports` instead of bisect-inserting; which is a followup for later

tybug: hmm, fair, I guess you don't lose anything if you simply compute relative to the worker and inserting at the right location

PR #107, opened 2025-05-16
Improvements

tybug:
I tried to add coverage stability here as well (re-executing to ensure stability, not the dashboard stability number), but it's too big of a feature for me to rush right now, so I dropped it.

Change descriptions are mostly in the commit messages


tybug:
right, forgot this depends on https://github.com/HypothesisWorks/hypothesis/pull/4392

tests/common.py
```diff
@@ -122,7 +122,7 @@ def fuzz(*, n=1, dashboard=False, test_path=None):
         "--dashboard" if dashboard else "--no-dashboard",
     ]
     if test_path is not None:
-        args += ["--", str(test_path)]
+        args += ["--", str(test_path), *(pytest_args or [])]
```
Zac-HD, line 125: Could simplify by defaulting to `pytest_args=()` above

src/hypofuzz/hypofuzz.py
```diff
@@ -100,6 +104,14 @@ def _should_trace(self) -> bool:
         return False
 
 
+# like choices_size, but handles ChoiceTemplate
+def _choices_size(choices: tuple[Union[ChoiceT, ChoiceTemplate], ...]) -> int:
+    return sum(
+        1 if isinstance(choice, ChoiceTemplate) else choices_size([choice])
+        for choice in choices
+    )
```
Zac-HD, line 112: Maybe we should have something like that upstream in Hypothesis?  Certainly doesn't block this PR though.

src/hypofuzz/hypofuzz.py
```diff
@@ -199,21 +211,50 @@ def __init__(
...
+        # replay.
+        self._replay_queue.update(self.db.fetch_corpus(self.database_key))
+        if not self._replay_queue:
+            # if no worker has ever worked on this test before, save an initial
+            # GENERATE report, so the graph displays a point at (0, 0).
+            # This is a bit of a hack. It might be better to do this in the
+            # dashboard plotting code instead, by adding a virtual point at (0, 0).
+            self.phase = Phase.GENERATE
+            self._save_report(self._report)
+            self.phase = None
```
Zac-HD, line 234: This does seem like a frontend concern, probably can be mostly handled on the dashboard-server side in linearization logic, and in the browser-side plotting code.  Feels like a followup though.

src/hypofuzz/hypofuzz.py
```diff
@@ -199,21 +211,50 @@ def __init__(
...
+
+        self._replay_queue.add((ChoiceTemplate(type="simplest", count=None),))
         for shrunk in [True, False]:
-            self._failure_queue.extend(
+            self._failure_queue.update(
                 self.db.fetch_failures(self.database_key, shrunk=shrunk)
             )
```
Zac-HD, line 240: Hmm, I think we want the failure_queue to replay shrunk failures, then unshrunk failures.  This suggests that we might go back to a generic `_replay_queue`, but entries are (ReplayPriority, nodes) with the appropriate (shrunk, unshrunk, covering) enum ordering?

src/hypofuzz/hypofuzz.py
```diff
@@ -511,6 +555,20 @@ def fuzz_several(*targets_: FuzzProcess, random_seed: Optional[int] = None) -> N
     for target in targets:
         target.startup()
 
+    dispatch = {}
+    for target in targets:
+        # each node only gets one fuzz process
+        assert target.database_key not in dispatch
+        dispatch[target.database_key] = target
```
Zac-HD, line 562: I Regret To Inform You, this is usually but not strictly always true (eg iirc if you set the `__name__` of a function Pytest probably still mostly works??).

I wouldn't bother if it was difficult to support, but accumulating a dict of lists and iterating over the callbacks seems equally easy to implement (`if event := ...: for callback in dispatch.get(event.database_key, []): ...`), etc.

Zac-HD (approved):
Minor comments below; fix or ignore or track for later as you prefer 🚀

src/hypofuzz/hypofuzz.py
tybug: hmm, if you have *database key* collisions, I think you're going to have a bad time regardless! Shared corpus isn't so bad, but shared observations, failures, and reports is.

That said, as you say this is easy to support, and we can think about other ramifications of database key collisions later.

src/hypofuzz/hypofuzz.py
tybug: possibly, and similarly for `choices_key` etc. I'm wary of locking in these semantics in hypothesis, since nothing there needs this right now. (But maybe in the future)

src/hypofuzz/hypofuzz.py
tybug: Added a `shrunk, unshrunk` priority for `_failure_queue`. The two queues are used slightly differently (failure_queue: delete failures from db after running), and I wasn't happier with my attempted unifying refactoring, so I'm putting that off for the future

src/hypofuzz/hypofuzz.py
Zac-HD: Oops, I was thinking of `nodeid` collisions (hence mentioning pytest), but it's _also_ true that tests can have database collisions; we try to avoid it but it _can and does_ occasionally happen in practice.

You're right that this is not a great time, but I think we do ~okay just by making sure that we handle it as best we reasonably can in each place that comes up outside of the literal database collisions.

src/hypofuzz/hypofuzz.py
Zac-HD: I'm weakly against HypoFuzz deleting failures from the database, except during shrinking - the ratio of time saved to time lost seems smaller than the chance that we delete something which we would have wanted to replay, later.

Issue #101, opened 2025-05-11
Try `sortedcontainers-stubs` for typechecking


Zac-HD:
[sortedcontainers-stubs](https://github.com/h4l/sortedcontainers-stubs) grew out of an unmerged PR against `sortedcontainers`; and while I'd prefer inline upstream annotations so I could actually evaluate e.g. `SortedList[Report]`, being able to type-check it would still be nice.

Issue #99, opened 2025-05-11
Get the `explain` phase working + consider improvements


Zac-HD:
We have [this test for the explain phase](https://github.com/Zac-HD/hypofuzz/blob/511b1b8f1aff174d2262efcb77aefdcb4c834a13/tests/test_fuzz_process.py#L30-L49), and... currently it demonstrates that we are _not_ leaving the nice comments about which arguments can vary without making the test ever pass.  Oops!

We might later investigate turning up some timeouts, or enabling sub-argument parts as per https://github.com/HypothesisWorks/hypothesis/pull/3624, to make use of the more relaxed time constraints in fuzzing mode 🙂  Similarly, since we have a browser interface, we might look at fancier interfaces to the coverage analysis ("scrutineer") which are impractical in terminal output.  Those are beyond the scope of this issue, though.

Zac-HD:
#6 covers the parts not implemented in #109 🎉

PR #109, opened 2025-05-20
Add tyche-style interface

### LINKED ISSUE #93: Check for `no_debug_ranges`
https://github.com/Zac-HD/hypofuzz/pull/92#discussion_r2072423444

We rely on [`code.co_positions`](https://docs.python.org/3/reference/datamodel.html#codeobject.co_positions) for locations (file, line, column) for branch events. These attributes are not present if `-X no_debug_ranges` is set.

I don't think there's a centralized location to check for this variable. The two ways I know of to set it are with `-X no_debug_ranges` (can check [`sys._xoptions`](https://docs.python.org/3/library/sys.html#sys._xoptions) (cpython specific)), and the [`PYTHONDEBUGRANGES`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONNODEBUGRANGES) environ variable (check os.environ).

Alternatively or additionally: run `code.co_positions()` on something we *know* should have position information, and warn/error if it doesn't.

tybug:
also closes #93 and closes https://github.com/Zac-HD/hypofuzz/issues/101 and probably #99 

- [ ] depends on https://github.com/HypothesisWorks/hypothesis/pull/4399


src/hypofuzz/database.py
```diff
@@ -404,6 +401,10 @@ def save_observation(
             obs_buffer.extend(
                 sorted(
                     list(self.fetch_observations(key)),
+                    # TODO run_start is the same for all reports from a worker,
+                    # since we use one StateForActualGivenExecution per FuzzProcess.
+                    # should we add a new `timestamp` hypofuzz-specific metadata?
+                    # (see also DataProvider code which relies on run_start).
```
Zac-HD, line 407: Somewhere I proposed overwriting the `run_start` with the `timestamp` from Hypofuzz; the former isn't really well-defined in the fuzzing setting.

Zac-HD:
PR looks good overall.  I wouldn't mind merging soon to make incremental review of followups easier.

Notes on the dashboard:

- We should have a fingerprint column in the summary table, hovertext "Number of distinct 'fingerprints'; covering corpus size".  We can do this even before we have true fingerprints on the backend.
  - bonus / later: toggle to switch the y-axis between branches and fingerprints
- We really need to show interactive draws, not just the base arguments to `@given()`.  That makes [this summary](https://hypofuzz.com/preview-109/example-dashboard/#/tests/tests%2Ftest_linearize.py%3A%3Atest_linearize_decomposes_with_addition) much less useful, for example.  I think if we pull the textual view out of observability that includes them?
- The "minimal covering examples" card should be an element of the "observations" card, and follow the covering / rolling toggle - currently there's no way to see the rolling-buffer inputs on the dashboard.
  - [maybe later] [eg here](https://hypofuzz.com/preview-109/example-dashboard/#/tests/tests%2Ftest_provider.py%3A%3Atest_changed_kwargs_pops_if_still_permitted), listing many textually-identical inputs is not great; can we instead put a little `and N more` or `x N` pill next to each such duplicate and only show them once?
- [for later] we should put some thought into coming up with a good set of tests to show on the example dashboard, and a checklist/runbook for what we want to do when updating it.  I'd like a few dozen diverse tests, which show a variety of failures, sizes, collection issues, efficiencies, etc.; and run for a few hours over multiple sequential workers to show all that off.  that's a largish project though and _definitely_ not urgent.
- unsatisfying alignment on the mosaic plots, eg Screenshot 2025-05-21 at 12 45 38 AM

tybug:
Thoughts on merging this now and following up in subsequent PRs?

src/hypofuzz/hypofuzz.py
```diff
@@ -483,6 +493,14 @@ def callback(test_case: dict) -> None:
             # because it doesn't find a pytest item context (I didn't look further
             # than that).
             test_case["property"] = self.nodeid
+            # run_start is relative to StateForActualGivenExecution, which we
+            # re-use per FuzzProcess. Overwrite with the current timestamp, so that
+            # the e.g. dashboard can reliably use this for sorting observations.
+            test_case["run_start"] = time.time()
```
Zac-HD, line 499: code is fine, but in the comment - this is an _unreliable_ (but good enough) way to sort observations.

src/hypofuzz/hypofuzz.py
```diff
@@ -483,6 +493,14 @@ def callback(test_case: dict) -> None:
...
+            # "arguments" duplicates part of the call repr in "representation".
+            # We don't use this for anything, so drop it.
+            test_case["arguments"] = {}
+            test_case = convert_to_fuzzjson(test_case)
```
Zac-HD, line 503: weak preference to keep this; it's a smallish part of the payload, and useful to expose for programmatic analysis via e.g. our dashboard-state endpoints (and maybe more endpoints in future).

Zac-HD (approved):
> Thoughts on merging this now and following up in subsequent PRs?

SGTM - it's a clear improvement on status quo.  See quick comments below and maybe the new Hypothesis explain observations, then let's merge!

src/hypofuzz/hypofuzz.py
tybug: `arguments` is actually half of the entire observation size, in the limit! (`call_repr` is the other half). Observations grow large because of arg reprs and pretty much nothing else.

We may eventually go the other way: drop `call_repr` from stored observations, and reconstruct dynamically from `arguments` on the frontend.

tybug:
(fyi, I also added sortedcontainers in the latest push; feel free to comment changes and I will follow up)

PR #113, opened 2025-05-31
Refactor execution as a `HypofuzzProvider` backend

tybug:
- [ ] depends on https://github.com/HypothesisWorks/hypothesis/pull/4422

The following now works:

```python
from hypothesis import given, strategies as st, settings

@given(st.integers())
@settings(backend="hypofuzz")
def f(n):
    print("call", n)

f()
```

### DIFF
--- a/src/hypofuzz/provider.py
+++ b/src/hypofuzz/provider.py
@@ -1,21 +1,75 @@
+import contextlib
+import inspect
 import math
+import os
+import platform
+import socket
+import subprocess
+import sys
+import time
+from base64 import b64encode
+from collections.abc import Generator, Set
+from contextlib import contextmanager
+from dataclasses import dataclass
+from enum import IntEnum
+from functools import lru_cache
+from pathlib import Path
 from random import Random
-from typing import Optional, cast
+from typing import Any, Optional, TypeVar, Union, cast
+from uuid import uuid4
 
+import hypothesis
+import hypothesis.internal.observability
+from hypothesis import settings
+from hypothesis.control import current_build_context
+from hypothesis.database import BackgroundWriteDatabase
 from hypothesis.internal.conjecture.choice import (
     ChoiceConstraintsT,
     ChoiceT,
+    ChoiceTemplate,
     ChoiceTypeT,
     choice_permitted,
+    choices_size,
 )
-from hypothesis.internal.conjecture.data import ConjectureData
+from hypothesis.internal.conjecture.data import ConjectureData, Status
 from hypothesis.internal.conjecture.providers import (
     COLLECTION_DEFAULT_MAX_SIZE,
     PrimitiveProvider,
 )
+from hypothesis.internal.escalation import current_pytest_item
 from hypothesis.internal.intervalsets import IntervalSet
+from hypothesis.internal.observability import InfoObservation, TestCaseObservation
+from hypothesis.internal.reflection import (
+    function_digest,
+    get_pretty_function_description,
+)
+from sortedcontainers import SortedList
 
-from hypofuzz.database import ChoicesT
+import hypofuzz
+from hypofuzz.corpus import Behavior, Choices, Corpus
+from hypofuzz.coverage import Branch, CoverageCollector
+from hypofuzz.database import (
+    ChoicesT,
+    DatabaseEvent,
+    DatabaseEventKey,
+    HypofuzzDatabase,
+    Observation,
+    Phase,
+    Report,
+    StatusCounts,
+    WorkerIdentity,
+    test_keys_key,
+)
+from hypofuzz.mutator import BlackBoxMutator, CrossOverMutator, Mutator
+from hypofuzz.utils import lerp
+
+T = TypeVar("T")
+process_uuid = uuid4().hex
+
+assert time.get_clock_info("perf_counter").monotonic, (
+    "HypoFuzz relies on perf_counter being monotonic. This is guaranteed on "
+    "CPython. Please open an issue if you hit this assertion."
+)
 
 
 def fresh_choice(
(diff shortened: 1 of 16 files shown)


src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
-        self.index = 0
+        self.collector = collector
+        self._index = 0
+        self._replay_queue: SortedList[ReplayQueueElement] = SortedList(
+            key=lambda x: (x[0], _choices_size(x[1]))
+        )
+
+        # TODO: thread through a random instance from FuzzProcess here, or access
+        # the current BuildContext random?
+        self.random = Random()
```
Zac-HD, line 130: I don't think it's worth trying for determinism outside of our self-tests, and we don't currently have much of a use even there.  Let's add a `seed: int | None = None` argument if that comes up in a future PR, and delete the comment for now.

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+        self.ninputs = 0
```
Zac-HD, line 131: should this be a property derived from status_counts?  and maybe cached

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+        self.database_key: Optional[bytes] = None
+        self.corpus: Optional[Corpus] = None
+        self.db: Optional[HypofuzzDatabase] = None
+        # per-test-case state, reset at the beginning of each test case.
+        self._state: Optional[State] = None
+
+    def _startup(self) -> None:
+        assert (
+            settings().database is not None
+        ), "pointless to use HypofuzzProvider without a database"
```
Zac-HD, line 152: probably want an if/raise here, just in case someone runs their fuzzer with `python -O ...`

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+
+        self.worker_identity = worker_identity(
+            in_directory=Path(inspect.getfile(context.wrapped_test)).parent
+        )
+        self.corpus = Corpus(self.db, self.database_key)
+
+        # Report that we've started this fuzz target
+        self.db.save(test_keys_key, self.database_key)
+        # save the worker identity once at startup
+        self.db.save_worker_identity(self.database_key, self.worker_identity)
```
Zac-HD, line 182: I'd move these to after all the other database-loading, albeit still before we save the initial-report: if we crash before actually doing anything, cleaner if it looks like this run never occured.

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+                if (
+                    self.corpus is not None
+                    and Choices(event.value) in self.corpus.corpus
+                ):
+                    return
+                self._enqueue(ReplayPriority.COVERING, event.value)
+            # (should we also replay failures found by other workers? we may not
+            # want to stop early, since we could still find a different failure.
+            # but I guess the same logic would apply to the worker which found the
+            # failure..)
```
Zac-HD, line 245: Yes, we do want to enqueue failures found by other workers - but then _what to do with that_ is fairly complicated.  We don't want to spend much time duplicating shrinking when another worker is doing that; but likewise we don't want to abandon the test if that worker happens to drop out.

Let's open an issue to revisit this later.  Current proposal: add to queue, but also stop working on this test for a few minutes if the shrunk instance is smaller than this workers best known failing input (for that interesting origin), and then revisit.  Make the queue consumer smarter about dropping not-smallest failures instead of replaying them all.

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+        assert self.corpus is not None
+        assert self.db is not None
+
+        self.db.save_report(self.database_key, report)
+        self._last_saved_report_at = self.elapsed_time
+
+        # Having written the latest report, we can keep the database small
+        # by dropping the previous report unless it differs from the latest in
+        # an important way.
+        if self._last_report and not (
```
Zac-HD, line 275: for a followup PR: let's instead track `self._last_timed_report`, or another mechanism to avoid the status quo where we keep the last timed report made before each event report.

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+
+    def before_test_case(self) -> None:
+        if not self._started:
+            self._startup()
+
+        assert self.corpus is not None
+        replay_priority = None
+        extra_queue_data = None
+        while self._replay_queue:
+            (replay_priority, choices, extra_queue_data) = self._replay_queue.pop()
```
Zac-HD, line 336: `pop(0)`!

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+            # We stay in blackbox mode until we've generated 1000 consecutive examples
+            # without new coverage, and then switch to mutation.
+
+            # TODO: currently hard-coding a particular mutator; we want to do MOpt-style
+            # adaptive weighting of all the different mutators we could use.
+            # For now though, we'll just use a hardcoded swapover point
+            if self.since_new_branch < 1000 or self.random.random() < 0.05:
+                mutator: Mutator = BlackBoxMutator(self.corpus, self.random)
+            else:
+                mutator = CrossOverMutator(self.corpus, self.random)
```
Zac-HD, line 360: Let's use a less dumb / more-smoothly-adaptive option?  

e.g. "use crossover with probability `(ninputs - len(corpus)) / ninputs`"

eventually we'll definitely want MOpt-style estimators, but this is still a huge improvement for now

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+        assert self.corpus is not None
+        assert self.database_key is not None
+        assert self.db is not None
+
+        assert observation.type == "test_case"
+        # run_start is relative to StateForActualGivenExecution, which we
+        # re-use per FuzzProcess. Overwrite with the current timestamp for use
+        # in sorting observations. This is not perfectly reliable in a
+        # distributed setting, but is good enough.
+        observation.run_start = time.time()
```
Zac-HD, line 393: let's make this "as of the start" by subtracting the increment we'll add to self.elapsed_time later.

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+            # sequence from unshrunk to shrunk - and failures are rare +
+            # deletions cheap. Just try deleting from both.
+            self.db.delete_failure(self.database_key, self._state.choices, shrunk=True)
+            self.db.delete_failure(self.database_key, self._state.choices, shrunk=False)
+            failure_observation = self._state.extra_queue_data
+            assert failure_observation is not None
+            assert isinstance(failure_observation, Observation)
+            self.db.delete_failure_observation(
+                self.database_key, self._state.choices, failure_observation
+            )
```
Zac-HD, line 424: followup for a future PR: let's evict shrunk failures only if the observation timestamp is 8+ days ago

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+
+        behaviors: Set[Behavior] = self._state.branches | (
+            # include |event| and |target| as pseudo-branch behaviors.
+            # TODO this treats every distinct value of target and every event
+            # payload as a distinct behavior. We probably want to bucket `v` for
+            # target (but not event?)
+            {
+                # strip in case v == "", as is the case for e.g. event without payload
+                f"event:{k}:{v}".strip(":")
```
Zac-HD, line 433: let's use a ternary; we don't want to strip a trailing `:` from `v`

src/hypofuzz/provider.py
```diff
@@ -26,29 +80,409 @@ def fresh_choice(
...
+        # Decide here, so that runtime doesn't bias our choice of what to observe.
+        if self.phase is Phase.GENERATE and (
+            self.elapsed_time > self._last_observed + 1
+        ):
+            self.db.save_observation(
+                self.database_key,
+                Observation.from_hypothesis(observation),
+                discard_over=300,
+            )
+            self._last_observed = self.elapsed_time
```
Zac-HD, line 468: we'd need to decide whether to observe before we increment self.elapsed_time

src/hypofuzz/provider.py
```diff
@@ -136,3 +570,61 @@ def draw_bytes(
         choice = self._pop_choice("bytes", {"min_size": min_size, "max_size": max_size})
         assert isinstance(choice, bytes)
         return choice
+
+
+@lru_cache
```
Zac-HD, line 575: let's unbound this cache size, and worker_identity, because there could be quite a few directories and the dicts are cheap.

PR #116, opened 2025-06-01
Performance + Tyche UI improvements

tybug:
I ran `hypothesis fuzz` overnight and then ran the dashboard under pyinstrument, and found some easy wins.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+# poor man's makefile. We could just use make here,
+# but ./build.sh <args> matches hypothesis.
+
+set -o xtrace
+set -o errexit
+set -o nounset
+
+ROOT="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+
+if [ "$1" = "docs" ]; then
+    sphinx-build "$ROOT/src/hypofuzz/docs" "$ROOT/src/hypofuzz/frontend/public/docs"
+elif [ "$1" = "dashboard" ]; then
+    npm --prefix "$ROOT/src/hypofuzz/frontend" run build
+else
+    echo "Unknown build target: $1"
+    exit 1
+fi
--- a/src/hypofuzz/dashboard.py
+++ b/src/hypofuzz/dashboard.py
@@ -1,14 +1,13 @@
 """Live web dashboard for a fuzzing run."""
 
 import abc
-import dataclasses
 import itertools
 import json
 import math
 from collections import defaultdict
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Optional, TypeVar, Union
+from typing import Any, Optional, TypeVar
 
 import black
 import trio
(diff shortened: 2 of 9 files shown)


src/hypofuzz/dashboard.py
```diff
@@ -609,6 +629,9 @@ def send_nowait_from_anywhere(msg: ListenerEventT) -> None:
         key = fuzz_target.database_key
 
         rolling_observations = list(db.fetch_observations(key))
+        # TODO this runs choice_to_bytes in fetch_corpus and then choice_from_bytes
+        # in fetch_corpus_observation. We should add a as_bytes param to fetch_corpus
+        # for performance
         corpus_observations = [
```
tybug, line 635: haven't done this since it might be obsoleted by smarter dashboard startup

tybug:
pagination for textual representation can be viewed here: https://hypofuzz.com/preview-116/example-dashboard/#/tests/tests%2Ftest_provider.py%3A%3Atest_drawing_prefix_exactly

tybug:
@Zac-HD I'm going to get this in because it will substantially decrease diffs for imminent PRs; comments welcome even after merge etc. The only things I think you might have comments on are:

* The pagination UI
* The decision to limit events to 1k per test
* Dropping "retried draw from" features from tyche
* (maybe) the new log x axis

build.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# poor man's makefile. We could just use make here,
+# but ./build.sh <args> matches hypothesis.
```
Zac-HD, line 4: fyi: Hypothesis also has `make`, just as a very think wrapper around... `build.sh` 😅

build.sh
```diff
@@ -0,0 +1,19 @@
...
+set -o nounset
+
+ROOT="$(git -C "$(dirname "$0")" rev-parse --show-toplevel)"
+
+if [ "$1" = "docs" ]; then
+    sphinx-build "$ROOT/src/hypofuzz/docs" "$ROOT/src/hypofuzz/frontend/public/docs"
+elif [ "$1" = "dashboard" ]; then
+    npm --prefix "$ROOT/src/hypofuzz/frontend" run build
+else
+    echo "Unknown build target: $1"
```
Zac-HD, line 17: "... known targets are:"

src/hypofuzz/dashboard.py
```diff
@@ -247,31 +246,52 @@ def linear_elapsed_time(self, *, since: float = -math.inf) -> list[float]:
...
-    del test["linear_reports"]
-    test["reports_by_worker"] = {
-        worker_uuid: [report_for_websocket(report) for report in reports]
-        for worker_uuid, reports in test["reports_by_worker"].items()
+    # Limit to 1000 reports per test. If we're over 1000 total reports, sample
+    # evenly from each worker.
+    #
+    # As a temporary stopgap until a more intelligent algorithm, sample the
+    # entirety of the first 20 reports per worker, then `::step` after. This
+    # helps with (4).
```
Zac-HD, line 254: If we want to downsample a lot, can we keep new behaviors and most recent report (for each worker) only?

I'd really like to avoid downsampling this harshly, but I guess maybe we have to...

src/hypofuzz/frontend/src/components/Pagination.tsx
Zac-HD, line 1: pagination seems fine; eventually I'd like to offer deeplinks to specific test cases - shareable URLs are a really useful pattern for communicating with colleagues - which pagination might make tricker, but I'm sure we can work it out.

src/hypofuzz/frontend/src/tyche/Features.tsx
```diff
@@ -7,6 +7,9 @@ export function Features({ observations }: { observations: Observation[] }) {
 
   observations.forEach(obs => {
     for (const key of obs.features.keys()) {
+      if (key.startsWith("Retried draw from")) {
+        continue
+      }
       features.add(key)
     }
   })
```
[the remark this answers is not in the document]
Zac-HD: I think for the draw-retries we could have a compact little table, so that this looked something like:

> Test cases with >=1 retries | Strategy
> -- | --
> 47% | `text(characters(codec='utf-8')).filter(not_yet_in_unique_list)`
> 7% | `integers(0, 1114111).filter(not_yet_in_unique_list)`
> 2% | `integers(1024, 9521).filter(not_yet_in_unique_list)`

src/hypofuzz/frontend/src/tyche/Representation.tsx
```diff
@@ -1,18 +1,22 @@
...
 import { TycheSection } from "./TycheSection"
+import { Pagination } from "../components/Pagination"
 
 hljs.registerLanguage("python", python)
 
 interface Props {
   observations: Observation[]
 }
 
+const perPage = 30
```
Zac-HD, line 15: Maybe 100?

src/hypofuzz/dashboard.py
tybug: quick math: 300 bytes per report, 10 tests, 1000 fingerprints per test = 300 * 10 * 1000 = 3000000 bytes = 3 mb. (and that's up to 30 if you have 100 tests etc).

Agree we want more intelligent sampling here, and maybe we can reduce this after optimizing the report format (msgpack etc) or the d3 plotting code.

tybug:
> I think for the draw-retries we could have a compact little table, so that this looked something like:
> https://github.com/Zac-HD/hypofuzz/pull/116#discussion_r2125544412

yeah, this is one option I'm considering (or a similar graph). Some strategy reprs are *really* long though, since we're using the `unwrap_strategies` version which uses the full strategy repr and not the LazyStrategy repr from the `st.lists` or whatever function call.

src/hypofuzz/frontend/src/tyche/Representation.tsx
tybug: perPage would have to be dynamic based on observation text size if we did this...I'm on an m1 macbook pro and was getting ~500ms delay on 50 per page for larger reprs, it's still somewhat noticeable on e.g. https://hypofuzz.com/example-dashboard/#/tests/tests%2Ftest_provider.py%3A%3Atest_drawing_prefix_exactly. It's not even the syntax highlighting (I tested), putting that much text in the dom is just expensive. People are going to have even crazier lager reprs. Maybe we should truncate at some point and require clicking to expand?

Zac-HD: never mind, we can revisit later if a user complains.

Zac-HD:
scroll-sideways table cell, or click to expand/collapse beyond two lines?

PR #119, opened 2025-06-03
Send dashboard state incrementally

tybug:
This refactors the websocket to send incremental json messages, instead of one big json message. We still batch at some level for performance (e.g. we send all of a worker's reports in a single message) send websocket frames have overhead.

Since we're receiving data asyncly / out of order now, this required correctly implementing the insertion logic for out-of-order reports in `add_report`.

### DIFF
--- a/src/hypofuzz/dashboard/dashboard.py
+++ b/src/hypofuzz/dashboard/dashboard.py
@@ -4,8 +4,9 @@ import abc
 import json
 import math
 from collections import defaultdict
+from enum import IntEnum
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Literal, Optional
 
 import black
 import trio
@@ -41,6 +42,16 @@ COLLECTION_RESULT: Optional[CollectionResult] = None
 websockets: set["HypofuzzWebsocket"] = set()
 
 
+class DashboardEventType(IntEnum):
+    # minimize header frame overhead with a shared IntEnum definition between
+    # python and ts.
+    ADD_TESTS = 1
+    ADD_REPORTS = 2
+    ADD_ROLLING_OBSERVATIONS = 3
+    ADD_CORPUS_OBSERVATIONS = 4
+    SET_FAILURE = 5
+
+
 def test_for_websocket(test: Test) -> dict[str, Any]:
     # Limit to 1000 reports per test. If we're over 1000 total reports, sample
     # evenly from each worker.
@@ -58,8 +69,7 @@ def test_for_websocket(test: Test) -> dict[str, Any]:
     # (4) sample more tightly earlier on and less tightly later on, since there is
     #     more behavior and fingerprint growth early. This is especially important
     #     for log x axis graphs, which show more early intervals than late intervals
-    count_reports = sum(len(reports) for reports in test.reports_by_worker.values())
-    step = (count_reports // 1000) + 1
+    step = (len(test.linear_reports) // 1000) + 1
     return {
         "database_key": test.database_key,
         "nodeid": test.nodeid,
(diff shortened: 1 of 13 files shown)


src/hypofuzz/frontend/src/components/TestTable.tsx
```diff
@@ -189,7 +189,12 @@ export function TestTable({ tests, onFilterChange }: Props) {
         data={sortedTests}
         row={row}
         mobileRow={mobileRow}
-        getKey={test => test.database_key}
+        // we'd like to use
+        //   getKey={test => test.database_key}
+        // here, but test is not guaranteed to have a databse_key set (if e.g.
+        // ADD_REPORTS arrives before ADD_TESTS), and I'm more worried about
+        // accidentally choosing an overlapping fallback key than trying to squeeze
+        // out performance here.
```
Zac-HD, line 197: the nursery fix might handle this?  Packet reordering is still possible on the wire, but I think logical reordering would require something weird with reconnections that we're already assuming doesn't happen.

src/hypofuzz/utils.py
```diff
@@ -51,3 +53,30 @@ def value(self, value: T) -> None:
...
+) -> list[T]:
+    # merges k sorted lists in O(nlg(k)) time, where n is the total number of
+    # elements.
+    #
+    # NOTE: this implementation is *not* a stable sort, since the heap key
+    #   (key(l[0]), i, 0)
+    # falls back to i when key(l[0]) is equal.
+    result: list[T] = []
+    heap: list[tuple[Any, int, int]] = []
+    assert all(len(l) > 0 for l in lists)
```
Zac-HD, line 69: Instead of asserting that the lists are nonempty, let's just `lists = [ls for ls in lists if ls]`

tests/test_utils.py
```diff
@@ -0,0 +1,30 @@
...
+def test_k_way_merge(lists):
+    assert k_way_merge(lists) == sorted(sum(lists, []))
+
+
+@dataclass
+class A:
+    value: int
+
+
+def test_k_way_merge_key():
```
Zac-HD, line 18: Could just add `key=sampled_from([None, lambda x: x, lambda x: -x])` to the test above.

tests/test_linearize.py
```diff
@@ -65,10 +92,10 @@ def reports(
...
         # reports from the same worker always have monotonically increasing
         # coverage, timestamps, and elapsed_time
-        # (TODO: we don't actually rely on timestamps being ordered. But our logic
-        # for unordered reports isn't correct yet either (we drop them, when we
-        # should correctly recompute/reinsert). Drop the .map(sorted) for timestamps
-        # when we implement this correctly.
+        # (TODO: we shouldn't actually rely on timestamps being ordered. But
+        # test_single_worker fails without it. This might be a real failure
+        # or it might be an acceptable edge case. Drop the .map(sorted) for timestamps
+        # when we investigte and fix it / figure out which it is.
```
Zac-HD, line 98: I'd prefer to at least understand what's going on before we merge; OK to defer if it's a rare-seeming edge case but shipping a bug would be sad.

src/hypofuzz/dashboard/dashboard.py
```diff
@@ -41,6 +42,16 @@
...
+    ADD_TESTS = 1
+    ADD_REPORTS = 2
+    ADD_ROLLING_OBSERVATIONS = 3
+    ADD_CORPUS_OBSERVATIONS = 4
+    SET_FAILURE = 5
+
+
 def test_for_websocket(test: Test) -> dict[str, Any]:
     # Limit to 1000 reports per test. If we're over 1000 total reports, sample
     # evenly from each worker.
```
Zac-HD, line 57: Let's also adjust the slicing logic to ensure that the last few (20?  at least one) are also included - it'd be sad to miss a breakthrough just because of our display logic.

src/hypofuzz/dashboard/test.py
```diff
@@ -47,14 +47,15 @@ def __post_init__(self) -> None:
...
-        # out-of-order reports), but it will for performance, by minimizing the
-        # number of .bisect calls in the worker reports list.
-        for report in sorted(
-            itertools.chain.from_iterable(reports_by_worker.values()),
-            key=lambda r: r.elapsed_time,
+        #
+        # This sorting doesn't matter for correctness (once we correctly insert
+        # out-of-order reports), but it does for performance, by minimizing the
+        # number of .bisect calls in the worker reports list. (I have not profiled
+        # this).
```
Zac-HD, line 56: I'm not following here; how can we bisect if it's not sorted?

src/hypofuzz/dashboard/test.py
```diff
@@ -47,14 +47,15 @@ def __post_init__(self) -> None:
...
+        for report in k_way_merge(
+            list(reports_by_worker.values()), key=lambda r: r.timestamp
```
Zac-HD, line 58: always sort by n_inputs or elapsed time, never by timestamp - in the unlikely event they disagree, it's probably because the clock was adjusted.

src/hypofuzz/dashboard/test.py
```diff
@@ -103,33 +104,34 @@ def _check_invariants(self) -> None:
...
             #
-            # TODO instead of dropping this report, insert it into the lists at
-            # the appropriate index, and recompute the diff for the next
-            # report. That way the e.g. dashboard graph still gets the report.
-            return
+            # we expect reports to *usually* arrive in-order, which case the
+            # appropriate report to diff against is reports[-1].
+            reports_index = bisect_right(
+                reports, report.elapsed_time, key=lambda r: r.elapsed_time
+            )
```
Zac-HD, line 120: ... I now think that we should probably define a `fast_bisect_right` helper which checks the end, then the start, and _then_ bisects if necessary.  log-n -> constant in the average case is pretty good!

src/hypofuzz/database.py
```diff
@@ -287,12 +285,16 @@ def from_reports(
...
+            if last_worker_report is None
             else max(
-                report.timestamp, last_report.timestamp_monotonic + elapsed_time_diff
+                report.timestamp,
+                last_worker_report.timestamp_monotonic + elapsed_time_diff,
             )
         )
 
+        assert elapsed_time_diff >= 0.0
+        assert timestamp_monotonic >= 0.0
```
Zac-HD, line 296: nope, this can happen without a bug if the user set their system clock to 1969 for some reason.  I agree that this is very strange, but it's not completely out of the question that someone would fuzz like that...

src/hypofuzz/dashboard/dashboard.py
```diff
@@ -123,31 +133,72 @@ async def initial(self, tests: dict[str, Test]) -> None:
...
-                return
-            if header["key"] is DatabaseEventKey.REPORT:
-                assert isinstance(data, Report)
+        # should we be passing in some global nursery here instead? I think this
+        # will block any callers until all send_event tasks complete, right?
+        # which we don't necessarily care about (though it's less important than
+        # sending all the events here in parallel).
+        async with trio.open_nursery() as nursery:
+            # we start by sending all tests, which is the most important
+            # thing for the user to see first.
```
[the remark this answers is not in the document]
tybug: ah, I don't actually need a guarantee that the `ADD_TESTS` event arrives before `ADD_REPORTS`; the dashboard is robust to this either way. `ADD_TESTS` is just both small and contains the most important info, so it's nice if it arrives first, for ux. I guess if trio doesn't guarantee queueing order for `start_soon` then it doesn't really matter whether we call `start_soon` for the tests before `start_soon` for the reports.

src/hypofuzz/dashboard/test.py
tybug: This should be "by minimizing the work each .bisect call does"; we'll still have to bisect, but if the bisect just appends to the end (as it will if we add reports in sorted order) then that's constant vs logn.

Though to your other comment, that's not the current performance characteristic of `bisect_right`; we should fix that.

src/hypofuzz/dashboard/test.py
tybug: thanks for continuing to remind me about this, it's not a conscious decision on my end, I just default to "timestamp" mentally as the sorting attr

src/hypofuzz/dashboard/dashboard.py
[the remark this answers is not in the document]
Zac-HD: I think the UX win is worth the tiny bit of extra latency from reduced parallelism?

(overkill mode: get send_event to support nursery.start, so you can ensure that we've started sending tests before we start_soon the various reports to be sent.  not worth it imo)

src/hypofuzz/database.py
tybug: agh, yup.

Zac-HD (approved):

Zac-HD:
(improvement on the status quo, so feel free to merge as-is or with more of the comments addressed)

src/hypofuzz/dashboard/dashboard.py
tybug: first and last 10 are now kept, and those 20 total don't count towards the soft limit of 1000

tests/test_linearize.py
tybug: whew, was a problem with the test itself (sorting by .timestamp instead of .elapsed_time).

PR #120, opened 2025-06-04
Send observation without reprs first, then reprs

Liam-DeVoe:
This doesn't buy us a whole lot right now, since the most prominent tyche chart is currently the mosaic of unique/duplicate, which relies on the repr. This does let other feature charts show earlier, though. And we might in the future precompute the unique/duplicate count on the backend.

This uses `run_start` as the primary key to identify an observation, which makes me nervous but seems to be fine in practice?


src/hypofuzz/dashboard/dashboard.py
```diff
@@ -204,66 +185,102 @@ def __init__(self, websocket: WebSocket, nodeid: str) -> None:
...
+        ]:
+            await self.send_event(
+                {
+                    "type": DashboardEventType.SET_OBSERVATION_REPRS,
+                    "nodeid": self.nodeid,
+                    "observation_type": obs_type,  # type: ignore
+                    "representations": [
+                        {
+                            # using run_start as the observation primary key
+                            "run_start": obs.run_start,
```
Zac-HD, line 240: Makes me nervous; [some clocks have pretty low resolution](https://peps.python.org/pep-0564/#annex-clocks-resolution-in-python).  Seems safe enough if we add the worker_uuid though.

tybug:
[drafting because I want to flesh out the dashboard pre-computation more, and make sure it's viable to defer observations like this]

PR #121, opened 2025-06-04
Load initial db state in separate thread

tybug:
Dashboard now starts up before all data is loaded. Data is streamed to existing connections as it loads. We don't currently have an indicator for "still loading x of n tests", which we might want to put as a dismissible notification in the top-right on page load.

I think I've done the trio code correctly here? `trio.from_thread.run(update_websockets, trio_token=trio_token)`

### DIFF
--- a/src/hypofuzz/dashboard/dashboard.py
+++ b/src/hypofuzz/dashboard/dashboard.py
@@ -11,6 +11,7 @@ import black
 import trio
 from hypercorn.config import Config
 from hypercorn.trio import serve
+from hypothesis import settings
 from hypothesis.database import (
     ListenerEventT,
 )
@@ -36,12 +37,12 @@ from hypofuzz.dashboard.test import Test
 from hypofuzz.database import (
     DatabaseEvent,
     DatabaseEventKey,
+    HypofuzzDatabase,
     HypofuzzEncoder,
     Observation,
     ObservationStatus,
     Report,
     ReportWithDiff,
-    get_db,
 )
 from hypofuzz.hypofuzz import FuzzProcess
 from hypofuzz.interface import CollectionResult
@@ -57,6 +58,7 @@ TESTS_BY_KEY: dict[bytes, "Test"] = {}
 LOADING_STATE: dict[bytes, bool] = {}
 COLLECTION_RESULT: Optional[CollectionResult] = None
 websockets: set["HypofuzzWebsocket"] = set()
+db: Optional[HypofuzzDatabase] = None
 
 
 def _sample_reports(
@@ -313,6 +315,8 @@ class TestWebsocket(HypofuzzWebsocket):
 
 
 async def websocket(websocket: WebSocket) -> None:
+    assert db is not None
+
     nodeid = websocket.query_params.get("nodeid")
     if nodeid is not None and nodeid not in TESTS:
         # requesting a test page that doesn't exist
@@ -543,7 +547,8 @@ async def handle_event(receive_channel: MemoryReceiveChannel[ListenerEventT]) ->
 
 
 def get_failure_observations(database_key: bytes) -> dict[str, Observation]:
-    db = get_db()
+    assert db is not None
+
     failure_observations = {}
     for maybe_observed_choices in (
         *sorted(db.fetch_failures(database_key, shrunk=True), key=len),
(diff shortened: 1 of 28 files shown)


src/hypofuzz/dashboard/dashboard.py
```diff
@@ -559,6 +547,37 @@ def send_nowait_from_anywhere(msg: ListenerEventT) -> None:
...
+            send_channel.send_nowait(msg)
+
+    db = get_db()
+    thread = threading.Thread(
+        target=_load_initial_state, kwargs={"trio_token": trio_token}
+    )
+    thread.start()
     # Any database events that get submitted while we're computing initial state
     # won't be displayed until a dashboard restart. We could solve this by adding the
     # listener first, then storing reports in a queue, to be resolved after
```
Zac-HD, line 583: We actually can't solve this in full generality (we don't have a globally consistent view), but we _can_ ensure that the events lost are only while we're loading state for _this particular_ test.  

Basically: start the listener _before_ loading initial state, but only keep reported changes from after that test finishes loading.  Any changes from before the test starts loading will be picked up by the initial load anyway, so it's only an issue for cases where something is written-but-not-read _during_ the load.

src/hypofuzz/dashboard/dashboard.py
```diff
@@ -559,6 +547,37 @@ def send_nowait_from_anywhere(msg: ListenerEventT) -> None:
         )
         TESTS[fuzz_target.nodeid] = test
 
+        async def update_websockets() -> None:
+            # TODO: make this more granular, so we send incremental batches
+            # of reports as they're loaded, etc.
+            for websocket in websockets.copy():
+                await websocket.initial({test.nodeid: test})
```
Zac-HD, line 554: Sending to everything seems super inefficient; can't we look up the specific websocket for this test and send it once?  Plus maybe to the overview page, I guess, but O(2) vs O(n) is a big deal on the inside of a size-n loop.

Zac-HD (commented):
- Trio code looks fine
- We'll need to document upstream that `ExampleDatabase` implementations must be threadsafe, and the the function passed to `.add_listener()` might be called from another thread.
- Relatedly we should have tests for database thread-safety for our 1P database implementations, and perhaps an example-database conformance test for use downstream...

src/hypofuzz/dashboard/dashboard.py
Zac-HD: If we change `websockets` from a set to a dict, with keys string-nodeid or None, I think we'd just drop this helper function and it'd be less code overall even if we make two calls to cover the overview + tests page.

src/hypofuzz/dashboard/dashboard.py
```diff
@@ -559,6 +549,37 @@ def send_nowait_from_anywhere(msg: ListenerEventT) -> None:
...
+                send_channel.send_nowait, msg, trio_token=trio_token
+            )
+        else:
+            send_channel.send_nowait(msg)
+
+    db = get_db()
+    thread = threading.Thread(
+        target=_load_initial_state, kwargs={"trio_token": trio_token}
+    )
+    thread.start()
```
Zac-HD, line 582: Looking at this more closely, starting a thread which we never join is a bit of a code smell.

What if we instead had a load_initial_state async function which loaded _one_ test and then started listening for updates, which internally used `trio.to_thread.run_sync` for the database interaction?  Then we could just dump them all into the same nursery which serves the webapp, and we'd get nice cleanup for everything.

tybug: so (pseudocode):
```
run_dashboard
    for test in test:
        nursery.start_soon(load_one_test)
    add_listener(listener)
```

with load_one_test setting some global state about when a test is currently being read (which our `listener` reads/respects), per your other comment.

(I just want to make sure you weren't intending each `load_one_test` to start its own listener; we should have one global listener per dashboard instance)

Zac-HD: yep, that's correct.

Then whatever is consuming our database-events channel will silently drop saves/deletes from tests which haven't finished loading yet, and process the rest in the ordinary way.

src/hypofuzz/dashboard/dashboard.py
tybug: I'd prefer to keep the filtering logic in the `Websocket` class, to support more complicated filtering (e.g. maybe a future "test directory" page, for tests only in a subdir). Maybe we could build an efficient `nodeid: list[websocket]` dispatch, but I'm tempted to defer this

Zac-HD (approved):

PR #123, opened 2025-06-06
Properly send failure deletion events

tybug:
I'm pretty happy with this, except again the note that the listener knows that a failure (observation) has been deleted, but not *which* failure. We don't expect there to be many failures, so we just rescan the the failures key for all tests which currently have a failure, and send the diff (tests which used to have a failure and now don't) to the dashboard.

### DIFF
--- a/src/hypofuzz/dashboard/dashboard.py
+++ b/src/hypofuzz/dashboard/dashboard.py
@@ -45,7 +45,13 @@ from hypofuzz.database import (
 )
 from hypofuzz.interface import CollectionResult
 
+# these two test dicts always contain the same values, just with different access
+# keys
+
+# nodeid: Test
 TESTS: dict[str, "Test"] = {}
+# database_key: Test
+TESTS_BY_KEY: dict[bytes, "Test"] = {}
 COLLECTION_RESULT: Optional[CollectionResult] = None
 websockets: set["HypofuzzWebsocket"] = set()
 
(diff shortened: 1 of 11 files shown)


src/hypofuzz/dashboard/dashboard.py
```diff
@@ -454,31 +479,70 @@ async def handle_event(receive_channel: MemoryReceiveChannel[ListenerEventT]) ->
...
+                        test.database_key_bytes
+                    )
+                    if not failure_observations.values():
+                        previous_failure = test.failure
+                        test.failure = None
+                        await broadcast_event(
+                            "delete",
+                            DatabaseEventKey.FAILURE_OBSERVATION,
+                            previous_failure,
+                        )
```
Zac-HD, line 525: I think we can do a lot better than a full scan here!  The `event.database_key` includes the db key of the failing test, and we should be find the Test instance with a dict lookup at that point.  Also if the DB _does_ support value deletion, we're not yet using that here.

tybug: damn, I even had a moment of "it sure would be nice to know the database key here, but it's just not possible in general", except it totally is lol.

and yeah there's a bunch of places we can do better with value deletion. I don't think I'm going to attempt them yet, because I don't have a good way to test it, so the code is likely to rot. I'll open an issue to track it

tybug:
test failure on previous push was a real one; a standard database key can of course have `b"."` at any location. You may be interested in taking a look at the new key length logic, which relies a bit on hypothesis internals.

src/hypofuzz/dashboard/dashboard.py
```diff
@@ -172,10 +178,23 @@ async def on_event(
...
                     "type": DashboardEventType.SET_FAILURE,
+                    "nodeid": value.property,
                     "failure": dashboard_observation(value),
                 }
                 await self.send_event(event)
 
+        if event_type == "delete":
+            # TODO when we support multiple failures, we'll need to send the
+            # specific observation that was deleted here, either via a
+            # DELETE_OBSERVATION or a SET_FAILURES (not the plural)
```
Zac-HD, line 189:
```suggestion
            # DELETE_OBSERVATION or a SET_FAILURES (note the plural)
```

PR #125, opened 2025-06-07
A few smaller (but still important) changes

Liam-DeVoe:


src/hypofuzz/mutator.py
```diff
@@ -36,7 +36,21 @@ def _get_weights(self) -> list[float]:
...
+                        for behavior in fingerprint
+                    )
+                )
+                # it's possible for a fingerprint to be empty. If it is, we expect
+                # that to be the *only fingerprint in the corpus, so it doesn't
+                # matter what weight we assign to it (but we can't error by giving
+                # an empty iterable to `min`, or dividing by zero with `total`.)
+                if fingerprint
+                else 1
+            )
```
Zac-HD, line 53: I'd pull this out into an inner function; I think it'd be a lot clearer with statemnts vs these nested expressions.

Zac-HD (approved):
👍

PR #129, opened 2025-06-08
Add selection ability to tyche mosaic charts

tybug:
Allow clicking on mosaic chart sections and headers to filter representations to those for the selected cells. I need to figure out a better way to communicate that selecting mosaic chart cells has an effect on the display of other tyche sections.

I think we'll want to extend this to filtering down other features as well. I think this could enable really powerful exploratory work ("what do the features with duplicate reprs and my_feature=2 look like?"). I'm not sure what the best UX is though.

### DIFF
--- a/src/hypofuzz/frontend/package-lock.json
+++ b/src/hypofuzz/frontend/package-lock.json
@@ -13,6 +13,7 @@
         "@types/highlight.js": "^9.12.4",
         "d3": "^7.8.5",
         "highlight.js": "^11.11.1",
+        "immutable": "^5.1.2",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "react-router-dom": "^6.22.0",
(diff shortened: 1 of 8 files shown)


tybug:
Drafting because I've sketched out a drill-down design that works over all features, and am confident I can implement a better UX than this PR

tybug:
This is still WIP, but it's mostly UI now, I think UX is roughly where I want it to be. This is already really cool to play around with.

To do:

- [ ] better ui for filter list at the top
- [ ] better filter names 
  - [ ] for nominal features
  - [ ] mosaic chart should special case entire-row or entire-col
- [ ] each individual feature / chart should list its currently active tag (in addition to the white border)? Unsure, might lose canonicalness. 

Wishlist (won't get to in this pr):

* [ ] each chart should show its current filtered state overlaid on top of its unfiltered state. Tricky because the proportions can change (especially for the mosaic chart, which has 2d transforms...), but very powerful because you can compare filtered to old state at a glance.

Zac-HD (approved):
Showing the currently-active filters is great!  I'd love the ability to click a little `(x)` to clear each filter pill too, non-blocking though.

Merge when ready 🫡

tybug:
there is a red `-` button on each pill which removes it 🤔

Zac-HD:
Ah, I'd read that as like the line in a plot legend.  dark-grey `x` to the upper-right of the pill is more obvious I think.

Issue #80, opened 2025-04-23
Frontend improvements for patches view


Zac-HD:
(just dropping some ideas here while I think about it)

- include a few sentences describing what this page is for, since it's a lot less obvious than the others.  We already have a note to improve Hypothesis' documentation (https://github.com/HypothesisWorks/hypothesis/issues/4309) and I think it'd be worth describing e.g. [this workflow](https://blog.nelhage.com/post/property-testing-like-afl/) (and [mindset](https://blog.nelhage.com/post/two-kinds-of-testing/)) in the Hypofuzz docs too.
- If there are only covering examples, don't show the "all" group on the page (we can assume that if there are failing examples, there must be covering examples too).  For each patch group shown, include clickable icons to copy-raw and to download the patch.
- Show a shell command which would apply the patch, `curl ... | git apply`, meaning we should have endpoints specific to all/failing/covering patches as a plain-text document.
- Syntax highlighting, e.g. https://highlightjs.org/demo

PR #140, opened 2025-06-12
Improve patches layout

Liam-DeVoe:
Addresses most of #80

- [x] depends on https://github.com/HypothesisWorks/hypothesis/pull/4437

* I don't love the idea of an "all" category; I think the ux for this is weird since it duplicates other categories, and I'm not sure how often people are copying both failing and covering examples in the same patch
* Performance of patch generation for large tests (>1k examples) is pretty bad, and freezes the dashboard process
  * need to investigate performance (unfortunately I think the answer is "libcst is slow") and/or push to a background thread

Still to do (probably later?):
* explanatory text on /patches
* copyable curl command (it's a lot of text, not sure where it fits on the ui. Maybe just an icon that represents "copy command", explaining on hover?)

### DIFF
--- a/src/hypofuzz/frontend/src/styles/styles.scss
+++ b/src/hypofuzz/frontend/src/styles/styles.scss
@@ -1271,7 +1271,7 @@ pre code.hljs {
 
         &--expanded {
             background: rgba(44, 62, 80, 0.03);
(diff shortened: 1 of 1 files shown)


Zac-HD:
online dashboard seems busted?

PR #149, opened 2025-06-16
Add "workers" ui page

tybug:
This is a pretty useful visualization of what your workers are actually doing. Improvements to come

thoughts welcome

followup todo:

- [x] clickable row to expand to display worker identity
- [ ] searchable box for worker identities? (what fields though?)

image


Zac-HD:
Intriguing!

This could obviously get pretty complicated though, as workers come and go over long periods of time (potentially with long gaps).  Might be worth writing up something about the UI/UX goals and tradeoffs so we have a record of why it's [whatever way it it ends up]?

I'd also like to make sure that the demo test run includes these aspects, so that users can see it working in a non-trivial case.

tybug:
yeah, we'll have to deal with gaps in the workers, as well as limiting the slider to a sane default (last 30 days? then click on a calendar for more).

I think we probably want a per-page page in the docs - which is also linked somewhere on each page? permanent bottom right open-book icon maybe? And for more complicated pages like Workers and Patches, an i-circle next to the header text which succinctly explains it.

Zac-HD:
> I think we probably want a per-page page in the docs - which is also linked somewhere on each page? permanent bottom right open-book icon maybe? And for more complicated pages like Workers and Patches, an i-circle next to the header text which succinctly explains it.

I worry about a docs page getting out of date; maybe we can toggle sufficient explanatory text in the frontend itself?  That'd also be nice for the online example dashboard.

tybug:
> I worry about a docs page getting out of date; 

fair, though I still think it's worth doing, and just taking care to keep it up to date. I've added a small amount of text for now (i-circle didn't turn out great)

Zac-HD:
looks like I can click on the worker details, but not on individual tests anymore.  Maybe we need a tab at one side to make sure that the worker details can be expanded, but also the tests remain clickable?

I'm also struck by the variety of date formats around the world; can we make that locale-responsive in the browser?  Low-prio.

tybug:
> looks like I can click on the worker details, but not on individual tests anymore. Maybe we need a tab at one side to make sure that the worker details can be expanded, but also the tests remain clickable?

the clicking is a bit buggy :/ but it does work about half the time

I intentionally left some padding on the left and right to ensure that you can always expand the worker even if the row is full of tests, but it's pretty tight. Maybe the *only* way to expand a worker should be a tab on the right

PR #154, opened 2025-06-16
Use a `distribute_nodes` function

tybug:
This doesn't actually incorporate estimator state right now. It just hardcodes an estimator of `1.0` for all nodes. But the groundwork and tests is all here for when we do plug in an estimator. (we weren't using estimator state before, so this doesn't change the status quo).

### DIFF
--- a/src/hypofuzz/bayes.py
+++ b/src/hypofuzz/bayes.py
@@ -1,0 +1,58 @@
+from collections.abc import Sequence
+from typing import Any
+
+
+def distribute_nodes(
+    nodeids: Sequence[str], estimators: Sequence[float], *, n: int
+) -> tuple[tuple[str, ...], ...]:
+    # We have x nodes node_i, each with an estimator \hat{v}_i for "behaviors per
+    # second". We have n bins (processes), and we want to distribute node_i
+    # into the n bins such that we maximize overall behaviors per second.
+    #
+    # If our estimators \hat{v}_i were static, then this is trivial: the
+    # overall behaviors per second is `sum from i=1 to c of max(p_i)`. Of course,
+    # our estimators are *not* static. Which means we are optimizing over something
+    # more complicated - a set of multi-armed bandit problems perhaps?
+    #
+    # A more naive quantity to maximize (minimize) is the largest bin sum.
+    # So we are minimizing `max(sum(p_i) for i in c)`. This is related to
+    # "makespan minimization", and is a classic bin packing problem. Optimality
+    # is NP complete, so we approximate the optimal solution with a greedy one,
+    # specifically "longest processing time first scheduling".
+    #
+    # (intuitively, we assign the "best" nodes to processes first. So with
+    # e.g. 10 processes with estimator of \hat{v}_i = 1 behavior per second (which
+    # is really good!) they would all go to different processes (at least until
+    # the cap of n processes), which is what we want.
+
+    assert len(nodeids) == len(estimators)
+    # estimators of 0 are mathematically valid, but semantically weird, and much
+    # more likely to be indicative of a logic error
+    assert all(estimator > 0 for estimator in estimators)
+
+    bins: list[list[Any]] = [[] for _ in range(n)]
+    nodes = [
+        {"nodeid": nodeid, "estimator": estimator}
+        for nodeid, estimator in zip(nodeids, estimators)
+    ]
+
+    # first, we sort the node_i in decreasing order by their estimator.
+    nodes.sort(key=lambda node: node["estimator"], reverse=True)  # type: ignore
+
+    # If we have fewer than `n` nodes, we repeat the list of nodes in decreasing
+    # order of their estimator until we reach `n` nodes. This ensures every
+    # processes receives a node (in fact, in this case, exactly one).
+    node_idx = 0
+    while len(nodes) < n:
+        nodes.append(nodes[node_idx])
+        node_idx = (node_idx + 1) % len(nodes)
+
+    # then, we assign each node_i to the partition with the least sum.
+    for node in nodes:
+        smallest_bin = min(
+            bins,
+            key=lambda bin: sum(node["estimator"] for node in bin),
+        )
+        smallest_bin.append(node)
+
+    return tuple(tuple(node["nodeid"] for node in bin) for bin in bins)
--- a/tests/test_bayes.py
+++ b/tests/test_bayes.py
@@ -1,0 +1,33 @@
+from hypothesis import given, strategies as st
+import itertools
+
+from hypofuzz.bayes import distribute_nodes
+
+
+@given(
+    st.lists(
+        st.tuples(
+            st.integers(0, 1000).map(lambda n: f"node{n}"),
+            st.floats(min_value=0, allow_infinity=False, exclude_min=True),
+        ),
+        min_size=1,
+    ),
+    st.integers(1, 100),
+)
+def test_distribute_nodes(items, n):
+    nodeids = [item[0] for item in items]
+    estimators = [item[1] for item in items]
+    partitions = distribute_nodes(nodeids, estimators, n=n)
+
+    total_nodes = sum(len(partition) for partition in partitions)
+    assert total_nodes == n if len(nodeids) < n else len(nodeids)
+    assert all(len(partition) > 0 for partition in partitions)
+
+
+@given(st.integers(4, 100))
+def test_distribute_nodes_more_processes_than_nodes(n):
+    # when there are more processes than nodes, we should see the highest value
+    # nodes getting assigned to processes first.
+    assert distribute_nodes(["1", "2", "3"], [1, 3, 2], n=n) == tuple(
+        itertools.islice(itertools.cycle([("2",), ("3",), ("1",)]), n)
+    )
--- a/src/hypofuzz/entrypoint.py
+++ b/src/hypofuzz/entrypoint.py
@@ -10,6 +10,8 @@ import hypothesis.extra.cli
 import psutil
 from hypothesis.internal.conjecture.providers import AVAILABLE_PROVIDERS
 
+from hypofuzz.bayes import distribute_nodes
+
 AVAILABLE_PROVIDERS["hypofuzz"] = "hypofuzz.provider.HypofuzzProvider"
 
 
@@ -155,24 +157,23 @@ def _fuzz_impl(n_processes: int, pytest_args: tuple[str, ...]) -> None:
         f"test{tests_s}{skipped_msg}"
     )
 
+    nodeids = [t.nodeid for t in tests]
+    # TODO actually save/load estimator state
+    estimators = [1.0 for _ in tests]
     if n_processes <= 1:
-        _fuzz(pytest_args=pytest_args, nodeids=[t.nodeid for t in tests])
+        _fuzz(pytest_args=pytest_args, nodeids=nodeids)
     else:
-        processes: list[Process] = []
-        for i in range(n_processes):
-            # Round-robin for large test suites; all-on-all for tiny, etc.
-            nodeids: set[str] = set()
-            for ix in range(n_processes):
-                nodeids.update(t.nodeid for t in tests[i + ix :: n_processes])
-                if len(nodeids) >= 10:  # enough to prioritize between
-                    break
-
-            p = Process(
+        partitions = distribute_nodes(nodeids, estimators, n=n_processes)
+        processes = [
+            Process(
                 target=_fuzz,
                 kwargs={"pytest_args": pytest_args, "nodeids": nodeids},
             )
+            for nodeids in partitions
+        ]
+        for p in processes:
             p.start()
-            processes.append(p)
+
         for p in processes:
             p.join()
 


Zac-HD:
Disagree.
- a global balance implies that all workers need to stop at the same time, which means we're going to have a period of idle time while n-1 workers wait for the slowest one to stop.
- like [crash-only software](https://en.wikipedia.org/wiki/Crash-only_software), why add a clean startup mechanism when the runtime mechanism you need anyway could do the job?

I'm imagining a hub-and-spoke design here, where each worker occasionally sends the 'hub' process a `dict[nodeid, fingerprints_per_second]` it currently has, and the hub replies with a `tuple[list[nodeid], list[nodeid]]` of tests to start/stop fuzzing (or maybe start/suspend/stop).  Once we've brought up all the estimators properly[^1], we should be able to balance value-of-compute across workers to within a few percent pretty easily.

It also has the substantial benefit that there's a straightforward path to extend this from balancing processes on a single host, to working via the database for an entire fleet 🙂 

[^1]: amortizing startup cost seems very important to me, and drops out of long-running workloads where this rebalancing matters anyway.

tybug:
> a global balance implies that all workers need to stop at the same time

not necessarily; each worker can check in with the hub (by reading from a shared `multiprocessing.Manager` dict or similar) for its new list of nodes whenever it would have switched targets anyway. We'll have the possibility of two workers fuzzing the same node simultaneously, but that's fine

> why add a clean startup mechanism when the runtime mechanism you need anyway could do the job?

the runtime mechanism might take a while to balance, and we have the estimators right there! might as well initialize a gradient descent with a good guess

I'm in agreement with the hub and spokes design, but I'm not convinced about "pick 3 and redistribute among those 3", when we could do "pick all n and redistribute among those n". I now think the hub should just continuously rebalance on an interval, rather than waiting for some condition

tybug:
haven't quite filled in all the estimator details yet, but this structure is what I was thinking of

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
+    while len(targets) < n:
+        # `targets` are in increasing order, so we index negatively to get
+        # a decreasing order
+        targets.append(targets[-target_idx])
+        target_idx = (target_idx + 1) % len(targets)
+
+    # then, we assign each target to the worker with the worst worker_behaviors.
+    # Since we're iterating over the targets in increasing order of behaviors
+    # per-second, adding a target to a worker will always increase its
+    # worker_behaviors.
```
Zac-HD, line 120: - prefer "lowest" or "highest" rather than "worst"
- I think we want to iterate in _decreasing_ order of _fingerprints_-per-second, adding to the current lowest-value worker, right?

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
+        worker_rates = e_worker_rates(
+            target_rates=[target.rates for target in worker["targets"]],
+        )
+        offset = 0.0
+        if target is not None and target.nodeid not in worker["current_worker"].nodeids:
+            # Add a penalty for switching nodes between workers. Since the ordering
+            # quantity is the e_worker_behaviors estimator of lifetime worker
+            # behaviors, we want to allow a node to switch workers if the ev
+            # differential is greater than the number of behaviors we expect to
+            # lose from spending time starting up this worker.
```
Zac-HD, line 134: checking: should `e_worker_behaviors` be called `e_worker_fingerprint_rate`?

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
+            # And the number of behaviors we expect to lose is the behaviors per
+            # second estimator for the worker, times the estimator for the startup
+            # time of this node.
+            #
+            # We are choosing the worker with the lowest score to add this node to,
+            # so if we want to encourage this node to be assigned to its current
+            # worker, we want that worker to have a low score, which means we
+            # want to increase the score of all other workers. So the offset here
+            # should be positive.
+            offset = worker_rates.per_second * target.e_startup_time
```
Zac-HD, line 145: Add `assert offset >= 0, offset`

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
+        e_lifetime = max(e_lifetime, DEFAULT_EXPECTED_LIFETIME_ESTIMATOR)
+        return (worker_rates.per_second * e_lifetime) + offset
+
+    for target in targets:
+        # find all the workers with the minimum value score, and randomly assign
+        # this target to one of them. Normally there won't be ties, and the target
+        # simply goes to the worst worker. But when fuzzing for the first time
+        # (or after a db wipe) where all targets have the same estimators, we
+        # don't want to end in an assignment where one worker is given n - 1 nodes
+        # and the other is given just 1.
```
Zac-HD, line 158: Hmm, my intuition is that this should be handled by our amortize-startup-cost logic (which iirc doesn't exist yet?) - maybe an elegant way to simulate that is to treat "not yet running" as a worker which it costs something to switch away from?

It seems to me that our desired outcome is to spend something like at most 10% of runtime on startup costs (after the first target), and enforcing that directly might mean that we can drop `DEFAULT_EXPECTED_LIFETIME_ESTIMATOR` entirely.

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
+        smallest_workers = _min_values(
+            workers,
+            key=lambda worker: worker_score(worker, target=target),
+        )
+        smallest_worker = random.choice(smallest_workers)
```
Zac-HD, line 163: I think we can actually just take `min()` here, and trust that the order of workers is approximately random.  The `worker_score()` should account for only-just-assigned targets, if it doesn't already.

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
-    # an estimator for the number of behaviors discovered per second, assuming
-    # one process is fuzzing this target continuously over that second.
-    # This is a simple adjustment of behaviors_per_input for the test runtime.
+    # per_second computation
     ninputs = target.provider.ninputs
     elapsed_time = target.provider.elapsed_time
 
     if elapsed_time == 0:
-        return 1
+        per_second = 1.0
```
Zac-HD, line 193: Do we have database state we can use to estimate this?

Even if not, I think the default `deadline` setting suggests that `10.` might be a better max-ignorance prior.

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
+
+
+def e_worker_lifetime(current_lifetime: float) -> float:
+    """
+    An estimator for the total lifetime of a worker.
+    """
+    # We use the doomsday-argument estimator that the total lifetime is twice the
+    # current lifetime. In the future, this could incorporate past worker
+    # lifetimes as well.
+    return current_lifetime * 2
```
Zac-HD, line 208: Let's incorporate the five-minute-minimum here instead, so that calling code doesn't need to think about that special case.

(it's also a nice place to put stuff like an eventual handler for spot-instance shutdown notification handlers)

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
-    inputs_per_second = ninputs / elapsed_time
-    return behaviors_per_input(target) * inputs_per_second
+
+def e_worker_rates(*, target_rates: Sequence[BehaviorRates]) -> BehaviorRates:
+    weights = bandit_weights(target_rates)
+    # the expected behavior rates of a worker is
+    # sum(probability * expected_value) for each of its targets.
+    # Note that this is tightly dependent on the sampling algorithm used in
+    # practice by the workers. If that changes (to e.g. thompson sampling), our
+    # estimators will need to change to use the same sampling algorithm as well.
```
Zac-HD, line 217: I don't think this is correct; if we change the `bandit_weights()` function, it's still fine because our estimator and our actor use the same weights.

src/hypofuzz/hypofuzz.py
```diff
@@ -588,18 +593,48 @@ def _rebalance(self) -> None:
...
+            worker_state = state["worker_state"]
+            worker_lifetime = worker_state["current_lifetime"]
+            for nodeid, rates in worker_state["nodeids"].items():
+                if nodeid not in targets:
+                    targets[nodeid] = (worker_lifetime, rates)
+
+                # if the nodeid already exists, but we have a better estimator
+                # for it, replace it
+                if worker_lifetime > targets[nodeid][0]:
+                    targets[nodeid] = (worker_lifetime, rates)
```
Zac-HD, line 620: this makes me think we should use a namedtuple or a dataclass; attribute accesswould be easier to read than indexing

src/hypofuzz/hypofuzz.py
```diff
@@ -588,18 +593,48 @@ def _rebalance(self) -> None:
...
+
+        # TODO estimate startup time of this target
+        # ("number of corpus elements" * "average input runtime" probably?)
+        targets = [
+            DistributeNodesTarget(nodeid=nodeid, rates=rates, e_startup_time=0)
```
Zac-HD, line 625: yes please 🙏

src/hypofuzz/hypofuzz.py
```diff
@@ -588,18 +593,48 @@ def _rebalance(self) -> None:
...
+        partitions = distribute_nodes(
+            targets,
+            n=self.n_processes,
+            current_workers=current_workers,
+        )
 
+        # communicate the worker's new nodeids back to the worker.
+        #
+        # the iteration order of the partitions returned by distribute_nodes is
+        # the same as the iteration order of current_workers
```
Zac-HD, line 637: Can we instead return a dictionary?  Or a list of current_workers which carry their respective partitions as attributes?  

(zipping just always makes me a little nervous + think that maybe we have the wrong data structure)

tests/test_bayes.py
```diff
@@ -1,7 +1,92 @@
...
+    st.lists(
+        st.builds(
+            DistributeNodesTarget,
+            nodeid=st.integers(0, 1000).map(lambda n: f"node{n}"),
+            rates=st.builds(
+                BehaviorRates,
+                per_input=st.floats(min_value=0, max_value=10**10, exclude_min=True),
+                per_second=st.floats(min_value=0, max_value=10**10, exclude_min=True),
+            ),
+            e_startup_time=st.floats(min_value=0, max_value=10**10, exclude_min=True),
```
Zac-HD, line 27: 10e6 is probably a more realistic upper bound; ten billion seconds or 0.1ns don't really make sense here.

tests/test_bayes.py
```diff
@@ -1,7 +1,92 @@
...
+        DistributeNodesTarget(
+            nodeid=str(i + 1),
+            rates=BehaviorRates(per_second=n, per_input=n),
+            e_startup_time=0.0,
+        )
+        for i, n in enumerate([1, 3, 2])
+    ]
+    assert set(distribute_nodes(targets, n=n)) == set(
+        itertools.islice(itertools.cycle([("2",), ("3",), ("1",)]), n)
+    )
```
Zac-HD, line 54: If this is testing ordering, asserting with a set is weird?

tests/test_bayes.py
```diff
@@ -1,7 +1,92 @@
...
+        (
+            _targets(("a", (1, 1), 0), ("b", (2, 2), 0), ("c", (3, 3), 0)),
+            1,
+            {("a", "b", "c")},
+        ),
+        (
+            _targets(("a", (1, 1), 0), ("b", (2, 2), 0), ("c", (3, 3), 0)),
+            2,
+            {("a", "c"), ("b",)},
+        ),
```
Zac-HD, line 80: Hmm, the greedy solution seems pretty bad here - don't we want `(a, b), (c,)`?

src/hypofuzz/bayes.py
tybug: No! `e_worker_behaviors` is the name for the inline estimator `worker_rates.per_second * e_lifetime` of the lifetime worker behaviors. Perhaps I should make this an explicit function for clarity.

src/hypofuzz/bayes.py
tybug: * +1
* fingerprints: yes (I'll refactor everything to make it agnostic between behaviors and fingerprints). lowest vs highest: I *think* these are equivalent algorithms. I had some reason in my mind for increasing rather than decreasing, but I forget it now.

src/hypofuzz/bayes.py
```diff
@@ -1,32 +1,224 @@
...
+    #
+    # Optionally, the current assignment `current_workers` of node ids to workers
+    # can be passed. This incorporates an overhead cost to switching a nodeid to a
+    # different worker. The algorithm is the standard bin packing algorithm, but
+    # with a penalty to a node being assigned to a worker other than its current
+    # worker.
+    #
+    # This penalty cost of switching a nodeid between workers is
+    # worker_behaviors_per_second * node_startup_cost_seconds, ie the number of
+    # behaviors we expect to lose by spending time starting up this node.
```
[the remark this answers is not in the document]
tybug: yup!

src/hypofuzz/bayes.py
tybug: nope, not equivalent: https://github.com/Zac-HD/hypofuzz/pull/154#discussion_r2174192667 is a counterexample. `3, (1, 2)` for highest -> lowest and `(3, 1), 2` for lowest -> highest.

src/hypofuzz/bayes.py
tybug: no database state yet; that'll come later with the once-per-worker estimator state

PR #155, opened 2025-06-16
Use softmax for multi-armed bandit

tybug:
I don't want to go all the way to thompson sampling yet, but softmax should be better than the current epsilon-greedy, especially when our estimators are as unstable as they are

Also improves our estimator to "behaviors per second" instead of "behaviors per input", and batches targets by 100 inputs instead of 1

### DIFF
--- a/src/hypofuzz/bayes.py
+++ b/src/hypofuzz/bayes.py
@@ -1,0 +1,31 @@
+import math
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from hypofuzz.hypofuzz import FuzzTarget
+
+
+def behaviors_per_input(target: "FuzzTarget") -> float:
+    # an estimator for the number of behaviors the next input will discover.
+    since = target.provider.since_new_branch
+    return (1 / since) if since > 0 else 1
+
+
+def behaviors_per_second(target: "FuzzTarget") -> float:
+    # an estimator for the number of behaviors discovered per second, assuming
+    # one process is fuzzing this target continuously over that second.
+    # This is a simple adjustment of behaviors_per_input for the test runtime.
+    ninputs = target.provider.ninputs
+    elapsed_time = target.provider.elapsed_time
+
+    if elapsed_time == 0:
+        return 1
+
+    inputs_per_second = ninputs / elapsed_time
+    return behaviors_per_input(target) * inputs_per_second
+
+
+def softmax(values: list[float]) -> list[float]:
+    # subtract max for numerical stability
+    softmaxed = [math.exp(value - max(values)) for value in values]
+    return [value / sum(softmaxed) for value in softmaxed]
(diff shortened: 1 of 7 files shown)


src/hypofuzz/bayes.py
```diff
@@ -0,0 +1,31 @@
...
+        return 1
+
+    inputs_per_second = ninputs / elapsed_time
+    return behaviors_per_input(target) * inputs_per_second
+
+
+def softmax(values: list[float]) -> list[float]:
+    # subtract max for numerical stability
+    softmaxed = [math.exp(value - max(values)) for value in values]
+    return [value / sum(softmaxed) for value in softmaxed]
```
Zac-HD, line 31: Cheaper to compute the max and the sum once, than per-element.

src/hypofuzz/bayes.py
```diff
@@ -0,0 +1,31 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from hypofuzz.hypofuzz import FuzzTarget
+
+
+def behaviors_per_input(target: "FuzzTarget") -> float:
+    # an estimator for the number of behaviors the next input will discover.
+    since = target.provider.since_new_branch
+    return (1 / since) if since > 0 else 1
```
Zac-HD, line 11: Function which takes one argument, which is a specific class?  Smells like a method.  (as does `behaviors_per_second`)

I'd also add a todo note for later that we might want to use a larger lookback window than last-one.  We also want to balance between using this-worker and linearized-history, but all that can come later.

src/hypofuzz/hypofuzz.py
```diff
@@ -289,32 +286,23 @@ def fuzz(self) -> None:
...
+                break
+
+            # choose the next target to fuzz with probability equal to the softmax
+            # of its estimator.
+            estimators = [behaviors_per_second(target) for target in self.targets]
+            estimators = softmax(estimators)
+            target = self.random.choices(self.targets, weights=estimators, k=1)[0]
+
+            for _ in range(100):
+                target.run_one()
```
Zac-HD, line 305: todo comment for a future PR: we could also scale this number up if we estimate a long time-to-bug or large gap to next-best option, to reduce switching overhead.  Similarly we could account for fixture costs in future.

tests/test_bayes.py
```diff
@@ -0,0 +1,21 @@
...
+    # check that softmax preserves ordering
+    for i in range(len(values) - 1):
+        if values[i] == values[i + 1]:
+            assert softmaxed[i] == softmaxed[i + 1]
+        if values[i] < values[i + 1]:
+            # note <= and not < here, since softmax may round down small
+            # differences in values to 0
+            assert softmaxed[i] <= softmaxed[i + 1]
+        if values[i] > values[i + 1]:
+            assert softmaxed[i] >= softmaxed[i + 1]
```
Zac-HD, line 21:
```python
idxs = [idx for idx, _ in enumerate(values)]
assert sorted(idxs, key=values.__getitem__) == sorted(idxs, key=softmaxed.__getitem__)
```
also assert that that sum(softmaxed) is approximately one.

src/hypofuzz/bayes.py
tybug: it does..but I want to keep all the estimator stuff in one place for now, to make it really clear when we're venturing into bayesian land

PR #156, opened 2025-06-17
Use per-file pytest collection

tybug:
This doesn't actually defer pytest collection yet, but it will make that diff easier to review!

### DIFF
--- a/src/hypofuzz/collection.py
+++ b/src/hypofuzz/collection.py
@@ -7,7 +7,8 @@ from collections.abc import Iterable
 from contextlib import redirect_stdout
 from dataclasses import dataclass
 from inspect import signature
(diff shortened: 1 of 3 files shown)


tybug:
hmm, that's probably the safe bet. It does seem a shame though! We'll be doing one collection per core (worker), and then a collection every time a worker gets rebalanced from the hub - even if the hub only added a single node to the worker.

Amortizing startup is definitely the bigger win, but some people have silly pytest collection times..

Zac-HD:
Oh, no, we can hold it down to one collection per worker at the start, and then rebalancing just creates the `FuzzTarget` wrapping the pytest item!

and yeah, some people have very high startup times, but that's on them / they can manually partition HypoFuzz too I guess.

PR #161, opened 2025-06-18
Try testing against pytest 6

tybug:
let's see how bad this will be to support; I'm tempted to drop it, pytest 7 was 3 years ago, and [download stats](https://pepy.tech/projects/pytest?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=line&versions=8.*%2C7.*%2C6.*) looks like about 5% usage


tybug:
alright, that CI failure is convincing enough for me; let's drop it

Issue #34, opened 2024-04-13
Setup hooks of _any_ sort


tarunik:
Right now, Hypofuzz provides no hooks for when it starts running a test target.  This is a problem because the only time prior to an individual fuzz run one can do setup work is at import time, which makes it nearly impossible to _not_ do that work when running the testsuite conventionally (i.e. with pytest and Hypothesis, not Hypofuzz).

At the very least, a hook called from FuzzProcess.startup would let me do Hypofuzz-only setup steps instead of having to try to detect whether we're running under pytest or not during import, which may not work anyway since Hypofuzz delegates test collection (and importing) to pytest.

tarunik:
Update: I was able to PoC this by manually monkeypatching FuzzProcess.startup, which'll get me going for now, but exposing a hook API of some sort would be appreciated.

tarunik:
In my case, I was trying to inject an instance of my unit-under-test into the fuzz testcase.  (Normally, this would be done with a pytest fixture, but those don't play nice with hypofuzz.)

Zac-HD:
Right, I think the right answer there is probably to support pytest fixtures, although I might leave function-scoped fixtures as an error.  That "just" takes setting them up and (for non-session-scoped) tearing them down when we switch between fuzz targets... it's conceptually pretty straightforward, just annoying to implement across various pytest versions.

PR #164, opened 2025-06-19
Add support for pytest fixtures

tybug:
closes https://github.com/Zac-HD/hypofuzz/issues/34

* I've left function-scoped fixtures as valid. I think this is good - if a user has explicitly suppressed `HealthCheck.function_scoped_fixture`, then I don't think hypofuzz should refuse to run function-scoped fixtures. 
* I'm still tearing down session-scoped fixtures every time we switch targets. This is a performance optimization, not a correctness one, so I'll work on it in a followup
  * in the same vein, if we switch to a target in the same (class/module/package), we don't have to tear down the fixture. session-scoped is the most impactful one though
  - followup: https://github.com/Zac-HD/hypofuzz/issues/184


src/hypofuzz/hypofuzz.py
```diff
@@ -100,40 +93,111 @@ def from_hypothesis_test(
...
+            name,
+            fixturedefs,
+        ) in self.fixtureinfo.name2fixturedefs.items():
+            # If there are multiple fixtures with this name, pick the last one,
+            # which is the closest one to us in scope.
+            #
+            # I'm not sure this is correct. More tightly scoped fixtures
+            # override less tightly scoped fixtures in pytest, but I don't know
+            # if this correctly handles interactions like dependent fixtures
+            # higher up the scope stack being overridden.
```
Zac-HD, line 182: Seems like a good case for some tests, which can check that pytest and hypofuzz see the same sequence of executed fixtures and of resulting values.

Zac-HD (approved):
I'd love to see some enhanced testing to make sure that we've got this right - and notice any regressions from e.g. pytest upgrades - but with that in mind, 🚀🚀🚀

src/hypofuzz/hypofuzz.py
tybug: I'm afraid of the correctness rabbit hole I'll find when I write tests for this 😅

tybug: Wrote one test that seems ok, though I'm still unconvinced.

PR #175, opened 2025-06-23
New graph display options for workers

Liam-DeVoe:
Related to https://github.com/Zac-HD/hypofuzz/issues/45

* single test: linearized / individual (color in age order) / latest
* landing page: linearized / latest

Todo (in a future pull):
* single test: sub-options for "individual": normal / offset x values (need to consider how this works with concurrent workers)
* both pages: "since" date selector for workers

Preview:

| together | separate | latest | 
|---|---|---|
| image | image | image |


Zac-HD:
- "together" doesn't look linearized in this image?  For showing separate-with-x-offset, I think we want to use the x-offset of the first generate-phase of that report; this means that the lines will have x-overlap and should be colored as for the current "separate" view.  That would also mostly obsolete the linearized view, not sure it's worth including both as options... we can just sort concurrent workers by the earliest report imo.

PR #177, opened 2025-06-25
Re-execute inputs for coverage stability

tybug:
Related: https://github.com/Zac-HD/hypofuzz/issues/5

I'm forging ahead here with a few opinionated design decisions:

* re-execute both new behaviors and "improved" (smaller choice sequences) behaviors
  * if the behaviors are the same across executions, add to the corpus
  * if the behaviors are different, drop the input entirely. Future work: add the stable subset to the corpus
* do not re-execute `Status.INTERESTING`, just report immediately
* for `elapsed_time` and `status_counts`, add an additional variable `elapsed_time_mutated` and `status_counts_mutated`, which tracks the number of "useful" inputs (ie freshly generated/mutated, not replayed). This is important now that we are doubling the number of executions for coverage inputs. They aren't persisted anywhere yet, but we'll use these for estimators in the future. See code comments for more of my thoughts here
* only activate Phase.REPLAY for the initial corpus replay, not for replaying e.g. database events from other live workers
  * one reason for doing this: switching phases on every replay generates lots of reports. could be improved by batching, but seems better to just track "number of useful inputs" separately from the phase

`corpus.add` has been pretty badly cannibalized, since the provider now needs more control over when things are added. RIP the nice interface.


src/hypofuzz/provider.py
```diff
@@ -134,29 +142,71 @@ def __init__(
...
+        # before any other enqueued elements, even if they get enqueued at a
+        # nominally higher priority before the corpus queue finishes.
+        #
+        # Essentially we have `ReplayPriority x [0, 1]`, which we could model
+        # by having both ${priority} and ${priority}_corpus and for each priority
+        # such that the corpus priorities are all higher, or we could model as
+        # two separate queues. This is the latter approach.
+        self._replay_queue: SortedList[QueueElement] = SortedList(
+            key=lambda x: (x[0], _choices_size(x[1]))
+        )
```
Zac-HD, line 191: Hmm, talk me through this reasoning?  I'd be inclined to use a single queue, on the basis that if another worker finds a failure (etc), then we likely want to replay that _before_ we finish initial load from the database - it's potentially a fairly substantial timesaver.  I also like the simplicity of having a single queue.

src/hypofuzz/provider.py
```diff
@@ -486,29 +541,116 @@ def after_test_case(self, observation: TestCaseObservation) -> None:
...
+            0 if new_fingerprint else self.since_new_fingerprint + 1
         )
-
-        if self.since_new_behavior == 0 or self.since_new_fingerprint == 0:
+        self.elapsed_time += elapsed_time
+        self.status_counts[status] += 1
+        if self._state.queue_priority is None:
+            # this was a "useful" / mutated / not-replayed-from-a-queue input.
+            self.elapsed_time_mutated += elapsed_time
+            self.status_counts_mutated[status] += 1
```
Zac-HD, line 631: I think here we actually want to distinguish between the blackbox "mutator" and the crossover mutator, since keeping those equal is the principled way to derive our `p_mutate`.

Zac-HD: (In the limit of really good heuristics or a custom-built generator, coverage-guided fuzzing is all cost and no benefit - we're almost certainly not there, but likely to be a lot closer than most fuzzers, and it'll be a nice signal to have for future research)

Zac-HD: There are some related ideas that I'd like to expose in some easy-to-understand way on the per-test page, maybe by showing the estimated progress curves for generation-only[^1], mutation-only, and adaptive-mixture approaches.  This will tend to over-rate mutation early on - because the seed corpus from generation is so important - but I don't really see a better option, unless we should maybe leave out mutation-only as unrealistic (or unobservable)?

My goal is to give users a sense of the effective time-multiple relative to generate-only as a function of campaign length, which helps inform the choice of how much to invest in fuzzing & when it's worth doing so at all.

We should generally expect that this shows Hypofuzz underperforms Hypothesis for [some shortish period], and then outperforms by a growing margin over time.

[^1]: note that this will me tracking, for each fingerprint, whether it's been found by generation, mutation, or both.  More metadata!

Zac-HD:
Also, it feels like we're within striking distance of #5 now, which I think we should treat as a fairly high priority - actionable feedback on coverage stability will really help users get a lot more out of fuzzing.

src/hypofuzz/corpus.py
```diff
@@ -203,53 +203,53 @@ def _evict_choices(self, choices: Choices) -> None:
...
+        if observation.metadata.data_status < Status.VALID:
+            return False
+
+        assert observation.metadata.choice_nodes is not None
+
+        new_behavior = any(
+            behavior not in self.behavior_counts for behavior in behaviors
+        )
+        new_fingerprint = behaviors not in self.fingerprints
+        new_coverage = new_behavior or new_fingerprint
```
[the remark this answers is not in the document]
tybug: I noticed this as well but thought being more explicit so we didn't have to think about this invariant was better, I'm indifferent though, happy to change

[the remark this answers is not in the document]
Zac-HD: I'd err on the side of minimal code, with a comment to explain this logic if you'd like it.

src/hypofuzz/provider.py
Zac-HD: I don't think it's double?  All failures can be the same priority whether from this or another process, and `STABILITY` can't come from another worker, so it's just `COVERING_LOCAL` plus `COVERING_REMOTE` I think.

src/hypofuzz/provider.py
tybug: > I think here we actually want to distinguish between the blackbox "mutator" and the crossover mutator, since keeping those equal is the principled way to derive our p_mutate

I think we want to track (whatever estimator state we use for estimating behaviors/fingerprints per second, currently "inputs since new coverage") independently for blackbox and mutator, and do a multi-armed bandit optimization between the two to generate `p_mutate`. This will cleanly inherit any thompson sampling improvements.

PR #193, opened 2025-07-16
Add `hypofuzz.detection.in_hypofuzz_run`, bucket target values

Liam-DeVoe:
We need some kind of bucketing, else eg `test_target_without_label` in the hypothesis test suite has ~infinite available branches. Ideally we would handle `target` scores specially in the mutator, but that's for later (and maybe much later, given how little-used target is).

### DIFF
--- a/src/hypofuzz/detection.py
+++ b/src/hypofuzz/detection.py
@@ -1,1 +1,5 @@
-in_hypofuzz_run = False
+#: |in_hypofuzz_run| is ``True`` when HypoFuzz has been invoked via
+#: ``hypothesis fuzz`` or another entrypoint, and ``False`` otherwise. You can
+#: use this to check whether HypoFuzz is currently fuzzing (or about to
+#: start fuzzing).
+in_hypofuzz_run: bool = False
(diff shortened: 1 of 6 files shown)


tybug:
(I forgot this branch already had a PR open; https://github.com/Zac-HD/hypofuzz/pull/193/commits/552afcab51c1cc6a28bb2dfe625b0599efe69217 is another change here)

src/hypofuzz/provider.py
```diff
@@ -133,6 +133,28 @@ class State:
...
 
+def bucket_target_value(v: Any) -> Any:
+    # technically, someone could do event("target", not_an_int) and get us
+    # confused with a target() call. We could/should track this provenance more
+    # carefully in observation.features
+    if not isinstance(v, (int, float)):
+        return v
+
+    # bucket by (base 2) orders of magnitude.
+    return int(math.log2(v))
```
Zac-HD, line 144: `OverflowError` for `v=1e999`, among other values.  

I think we'll need a more sophisticated approach here, possibly with dynamic bucketing... but for now I'd consider simply not treating targets as virtual branches, except for the current-largest value (breaking ties by choice-sequence sort order).

PR #197, opened 2025-07-17
Styling and slight wording change

tybug:
I'm surprised every time by how big of a difference adding icons makes. I don't dislike the diff or collection icons, but we might be able to do better.

image


Zac-HD:
I like it!  Maybe a map or binoculars icon for collection?  or a basket?

I'd also consider hiding the together/latest graph toggle, when there's only one worker in the history.

PR #199, opened 2025-07-19
Work on roadmap docs, adjust `in_hypofuzz_run`

tybug:


src/hypofuzz/docs/manual/collection.rst
```diff
@@ -1,18 +1,43 @@
...
+
+``hypothesis fuzz`` does not run pytest plugins during collection. If your standard pytest command includes plugin parameters like ``pytest -n auto`` (from e.g. :pypi:`pytest-xdist`), do not pass these parameters to ``hypothesis fuzz``.
 
-Support for other test runners, such as :mod:`python:unittest`, is on our roadmap.
 
 Skipping tests under HypoFuzz
 -----------------------------
 
-If you would like to skip a test under HypoFuzz while keeping it part of your normal test suite, you can use |in_hypofuzz_run|:
+If you would like to skip a test under HypoFuzz while keeping it part of your normal test suite, you can use :ref:`pytest:pytest.mark.skipif ref` in combination with |in_hypofuzz_run|:
```
Zac-HD, line 40: Code block below uses that as a boolean when it's now a function!

I'd also suggest showing the sys.modules trick, to avoid importing hypofuzz if it's not present, and document that under in_hypofuzz_run too.

tybug: I'd hope they don't need the sys.modules trick, since importing hypofuzz won't have side-effects on `in_hypofuzz_run`, so they can do so unconditionally!

tybug: sure, but that's just a generic perf comment, not hypofuzz-specific. I might mention it briefly but I wouldn't go into too much detail

PR #216, opened 2025-07-30
Re-execute rolling observations for coverage stability

tybug:
See https://github.com/Zac-HD/hypofuzz/issues/5.

Stability metric between 0 and 100 is now reported in the table. The metric is "stability of rolling observations". We've previously discussed reporting two stability metrics, the other being the stability of the covering corpus. This doesn't make sense as stated because we only add stable entries to the corpus to start. But we could report stability of entries which cover new branches. I'm not sure this is as or more valuable than the random sample of rolling observations here.

* If we want to replay an input for both observation stability and coverage stability, do both in the same execution. `QueuePriority.STABILITY` has been generalized to "any stability re-execution", with `extra_queue_data["reasons"]` directing what we do with the results.
* Fix `since_new_{behavior, fingerprint}` incrementing during replay, instead of just mutation

- [x] update preview dashboard data

image


tybug:
I think `scale-balanced` is a better fit than either `location-pin-lock` or `table-cells-row-lock` – though not perfect – but I do appreciate the overloaded semantics with law/justice. I've switched to `location-pin-lock` for now.

I'm starting observability stability re-execution overhead at a cap of 2%.

> maybe also a pill of some kind for unstable cases? 

I don't want to draw *too* much attention to it. I'll try a very light yellow/red pill, or possibly only tinting the text, without a pill.

PR #219, opened 2025-08-05
Save reports on worker switch + Tyche improvements

tybug:
Stability is a new tyche feature:

image

The gear icon indicates we generate this automatically, to avoid confusing the user ("why do I have a feature which doesn't correspond to any event/target/note call in my code?") 

image

I do not know why there are so many "unknown" stability events here. There should be 0. Probably a bug.


tybug:
ah, not a bug: we are actually hitting the prescribed 2% time limit on observation stability.

PR #229, opened 2025-08-27
Add a new example dashboard

Liam-DeVoe:
I'll let the example dashboard speak for itself here, since that's all users will see, not the code. If it can't stand on its own then it needs to be improved

noticed issues: /patches page doesn't work


Liam-DeVoe:
- dashed lines feels like an inversion, where we want to be drawing attention *to* failures, but dashed draws it away. It kinda works though since dashed implies "inactive", and coverage for failed tests won't change since it's not running anymore
- I like some kind of explosion marker or similar for failing tests, but I want to think about the right icon+style+color
- `test_json_roundtrip_fails_on_inf` having 0% coverage is real, sort of: its first and only observation really was unstable due to a runtime import of `typing-extensions`

